Issue #2, opened 2022-06-08
Roaming Status API Proposal


eric-murray:
A proposal for a Roaming Status API has been uploaded [here](https://github.com/camaraproject/WorkingGroups/blob/main/APIBacklog/documentation/Contributions/API%20Proposals/Roaming%20Status.md). Please add any comments you may have.

Unfortunately, I don't have permissions to add labels to this issue, so maybe somebody could do that for me.

MarkusKuemmerle:
Reopened in the DeviceStatus Sub Project

eric-murray:
I uploaded the original document to [here](https://github.com/camaraproject/DeviceStatus/blob/main/documentation/SupportingDocuments/Roaming%20API%20Proposal%20-%20Vodafone.md) and fixed the broken link in my original comment on this issue.

The idea is to discuss the use cases proposed in the document as a possible way to evolve the Device Status API, so I will keep this issue open for now.

NoelWirzius:
@eric-murray please close Issue

PR #5, opened 2022-10-01
Initial contribution of API spec check-device-connectivity v0.2.0

akoshunyadi:
Initial contribution of the API spec check-device-connectivity with the version 0.2.0. #2 
The API enables to query the roaming status of a UE. further extensions with other connectivity related functionalities are possible.

### DIFF
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -13,7 +13,7 @@ externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
 security:
-  - {}
+#  - {}
   - oAuth2ClientCredentials: []
 #  - BasicAuth: []
 #  - apiKey: []
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -1,0 +1,245 @@
+openapi: 3.0.3
+info:
+  title: Check device connectivity API
+  description: Service Enabling Network Function API for monitoring device connectivity
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.2.0
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  - {}
+  - oAuth2ClientCredentials: []
+  - BasicAuth: []
+  - apiKey: []
+  - three_legged:
+    - device-connectivity-read
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: connectivity/v0
+        description: Base path for Device Connectivity API
+tags:
+  - name: Device Connectivity Status
+    description: Operations to validate device connectivity status
+paths:
+   /status:
+    post:
+      tags:
+        - Device Connectivity Status
+      summary: 'Get the current connectivity status'
+      operationId: getStatus
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RequestStatus'
+        required: true
+      responses:
+        '200':
+          description: Contains information about current connectivity status
+          content:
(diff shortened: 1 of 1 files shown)


code/API_definitions/check-device-connectivity.yaml
```diff
@@ -0,0 +1,245 @@
...
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.2.0
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  - {}
```
eric-murray, line 16: My preference would be to remove the "no authentication" option, or at least comment that it is only for test / development purposes, but it can stand for now

eric-murray (approved):

SylMOREL (approved):

code/API_definitions/check-device-connectivity.yaml
bigludo7: Just a small detail but for consistency sake we should have the same security clause for all Camara API specificationr ?
the QoD API v0.8.0 has:
`security:
  - oAuth2ClientCredentials: []`

Perhaps we should aligh QoD security.

monamok:
After reviewing the current version included in this pull request with our Product and Network Engineering teams in Telefónica, we would like to suggest the following changes:

As the only functionality that it is included in this API is to check device roaming status we suggest to call the API Device Roaming Status with a single endpoint GET /roaming which returns a boolean for roaming: true/false. These are the reasons why we find this solution interesting:

* GET /roaming
  - Current input field that won't apply:
	- To identify the device we think that we have to move from the current **ueId** object to a more convenient identifier to avoid ambiguity in how the mobile connectivity of the device that is going to be identified is, and to deal properly with sensible information. There is a proposal for this change that is already being discussed within Commonalities, as it is transversal to almost every API: [WorkingGroups Issue #101]( https://github.com/camaraproject/WorkingGroups/issues/101)
	
	- For the same reason **uePort** wouldn't apply
	- If the only use case of this API is to identify if roaming status **eventType** and **eventStatus** could be both removed for simplicity
  - Output:
	- Current output parameters can be replaced by the following one:

```	
    RoamingStatusResponse:
      type: object
      required:
        - roaming
      properties:
        roaming:
          $ref: '#/components/schemas/ActiveRoaming'
    ActiveRoaming:
      description: Roaming status. True, if it is roaming
      type: boolean
```

* securitySchemes:
    - Only three_legged should be considered requiring user authentication and consent gathering

	
In order to have the solution visible and accesible I will create antoher PR so you can see our proposal. But meanwhile please feel free to share your comments.

bigludo7:
Hello
Regarding the API proposal name change, from my perspective, it is depending of this API planned roadmap & granularity. 

If in future release we want to extend this API (initial scope is check if a device is losing connection to the network or gets reachable again and the roaming status) then we should keep Device status name (and we have a v1 with only roaming). 

If we prefer to have several end points one for roaming and another for connection status then, yes, we can change the name.

monamok:
Hello
Thank you @bigludo7 for your comment. I totally agree with you. As you mentioned changing the name makes sense if check roaming status is the only functionality of this API.

The other part of the scope (to check if a device is losing connection to the network or gets reachable again) is not currently included in Telefonica roadmap and as we also believe that APIs should not be implemented partially, we prefer to have two separate APIs. It’s more practical and we avoid future problems. 

Do you consider this solution feasible? Do you find it aligned with your company’s roadmap?

bigludo7:
Hello @monamok 
Thanks for your answer. This is fine for me and providing only a resource checking roaming status could work.

monamok (commented):
Thank you @eric-murray for the comments. As version 0.2.0 of this API we can go ahead y accept it. But please keep in mind that for the first official version we need to make some changes mainly about these points:
-	I think we all agree that having a single _/status_ path is not the best way of defining the API. 
-	Telefonica policy does not allow the use of personal information as input nor output. But we can postpone the discussion till [WorkingGroups Issue #101]( https://github.com/camaraproject/WorkingGroups/issues/101) get resolved and there is a common approach.

shilpa-padgaonkar:
As agreed, merging the pull request so that we can have a base version. New issues should be created to bring in further improvements in the spec.

PR #12, opened 2022-12-15
Upload Roaming API Proposal - Vodafone.md

eric-murray:
Upload of original Roaming API proposal from Vodafone


bigludo7:
Hi @eric-murray 
Very interesting proposal.
I need to share this with colleagues and with end of year holidays starting tonight I'm afraid to not get back to you before early January.
Sorry for the delay.

eric-murray:
Hi @bigludo7

This is actually the proposal I originally submitted to the Backlog WG. Markus transferred the associated issue over to Device Status WG (now Issue #2) but did not copy the proposal document, so the issue now has a broken link. I'm just trying to fix that, but if you have any comments on the proposal, you can add them to the existing issue.

sachinvodafone (approved):

Issue #18, opened 2023-01-31
More Details for Roaming Status 


NoelWirzius:
**Summary**
===========
Expanding the roaming status with additional informations, which can help the developer to improve application. Example to set the right profiles for them.  

**Changes** 
==========
*Adding new return parameters 
  * Country Code 
    * Is the Code of the country where the device is 
  * Country Name  
    * Is the Name of the country where the device is

sachinvodafone:
I would recommend to have following in response apart from roaming status:

- "Home_PLMNid"
- "Visited_MNC"
- "Visited_MCC"

yoonsun3:
I think we can also offer roaming operator information from PLMN. But, it seems to be difficult to match and manage the Operator Code and Operator Name.

- Operator Code (MNC)
    - Is the Code of the operator where the device is
- Operator Name
    - Is the Name of the operator where the device is

eric-murray:
I would avoid returning home PLMN Id as part of the roaming report as that is really a different use case unrelated to roaming. There may be reasons why an API caller wants to know with which mobile operator a given subscriber (MSISDN) has a contract, but that is a different type of status, so should at least be a different path, if not a different API, as they should not have to call a "roaming status" path to get this information. It would also require separate user consent.

I would also avoid returning Operator Name. Knowing the visited PLMN Id would allow the API caller to determine this for themselves if they really wanted to know, but it would be difficult for the API provider to do this reliably and I think this would be an unnecessary  overhead.

yoonsun3:
I also agree that there will be no case to use the operator information. I just suggested that because it would be like to create a standard with optional parameters so that information can be returned, if possible.

akoshunyadi:
Do we have a common understanding, whether these 2 new fields (country name and code) should be included in the response also for the home network?

eric-murray:
@akoshunyadi 

Do you mean when a mobile is not roaming? For a German mobile subscription, for example:
```
{
    roaming: false
    countryCode: 262
    countryName: ["DE"]
}
```

I'd say that was up to the implementation but, knowing the identity of the UE, it is not hard to work out what country is the home country. But `countryCode` and `countryName` are both optional, so it is valid not to include them.

sachinvodafone:
I agreed with @eric-murray for not including countrycode and countryName if person is not in roaming. It is not relevant.

NoelWirzius:
Closing this issue

PR #21, opened 2023-02-07
Add new return parameters Country Code and Country Name

ravindrapalaskar17:
* Add new return parameters Country Code and Country Name 

More Details for Roaming Status #18 :- With reference to this issue on CAMARA we  added new return parameter 
1. Country Code
2. Country Name

### DIFF
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -165,11 +165,13 @@ components:
     CountryCode:
       description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
       type: integer
-      example: 262
+      example: 340
     CountryName:
-      description: The ISO 3166 ALPHA-2 code of the country.
-      type: string
-      example: "DE"
+      description: A list of ISO 3166 ALPHA-2 country-codes of mapped to the mobile country code (MCC).
+      type: array
+      items:
+        type: string
+      example: [BL,GF,GP,MF,MQ]
     RequestStatus:
       type: object
       properties:
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -165,9 +165,9 @@ components:
         - ROAMING_ON
         - ROAMING_OFF
     CountryCode:
-      description: The country calling code as an identifier for the country and the dependent areas.
-      type: string
-      example: "49"
+      description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
+      type: integer
+      example: 262
     CountryName:
       description: The ISO Code of the country.
       type: string
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.0
+  version: 0.2.1
 externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
@@ -30,7 +30,7 @@ servers:
         description: Base path for Device Connectivity API
 tags:
   - name: Device Connectivity Status
-    description: Operations to validate device connectivity status
+    description: Operations to validate device connectivity status and Country information.
 paths:
    /status:
     post:
@@ -99,10 +99,16 @@ components:
           $ref: "#/components/schemas/ConnectivityEventType"
         eventStatus:
           $ref: "#/components/schemas/EventStatusType"
+        countryCode:
+          $ref: "#/components/schemas/CountryCode"
+        countryName:
+          $ref: "#/components/schemas/CountryName"  
       required:
         - ueId
         - eventType
         - eventStatus
+        - countryCode
+        - countryName
     UeId:
       description: User equipment identifier
       type: object
@@ -158,6 +164,14 @@ components:
       enum:
         - ROAMING_ON
         - ROAMING_OFF
+    CountryCode:
+      description: The country calling code as an identifier for the country and the dependent areas.
+      type: string
+      example: "49"
+    CountryName:
+      description: The ISO Code of the country.
+      type: string
+      example: "DE"
     RequestStatus:
       type: object
       properties:
--- a/code/API_definitions/check-device-connectivity.yaml
+++ b/code/API_definitions/check-device-connectivity.yaml
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.1
+  version: 0.3.0
 externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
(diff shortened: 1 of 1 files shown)


code/API_definitions/check-device-connectivity.yaml
```diff
@@ -158,6 +164,14 @@ components:
       enum:
         - ROAMING_ON
         - ROAMING_OFF
+    CountryCode:
+      description: The country calling code as an identifier for the country and the dependent areas.
+      type: string
+      example: "49"
```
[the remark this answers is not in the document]
ravindrapalaskar17: Thanks for feedback we updated YAML and pushed commit in this PR.

monamok (commented):
Do these new fields always return value? What happens if "eventStatus" is "ROAMING_OFF"? It returns the country information where the mobile belongs to?

code/API_definitions/check-device-connectivity.yaml
```diff
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.0
+  version: 0.2.1
```
monamok, line 11: Do we add it as a fix or should we start with 0.3.0?

code/API_definitions/check-device-connectivity.yaml
```diff
@@ -99,10 +99,16 @@ components:
...
+        countryCode:
+          $ref: "#/components/schemas/CountryCode"
+        countryName:
+          $ref: "#/components/schemas/CountryName"  
       required:
         - ueId
         - eventType
         - eventStatus
+        - countryCode
+        - countryName
```
eric-murray, line 111: `countryCode` and `countryName` should not be `required`. Even if we agree that the home `countryCode` and `countryName` can be returned when the device is not roaming (this information can probably be deduced from `ueId` anyway), we need to consider user consent.

The user may consent to their roaming status being made available to the API caller, but not the country in which they are roaming. Providing the roaming country provides much more information on the device location than just knowing that the device is not in the home country.

code/API_definitions/check-device-connectivity.yaml
```diff
@@ -158,6 +164,14 @@ components:
       enum:
         - ROAMING_ON
         - ROAMING_OFF
+    CountryCode:
+      description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
+      type: integer
+      example: 262
+    CountryName:
+      description: The ISO Code of the country.
```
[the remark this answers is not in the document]
monamok: Another thing is that, it seems that in some cases there isn't a 1:1 relationship between: MCC and ISO Country code:

<html>
<body>
<span><span class="ui-provider xd b c d e f g h i j k l m n o p q r s t u v w x y z ab ac ae af ag ah ai aj ak" dir="ltr"><figure class="table">

340 | BL/GF/GP/MF/MQ
-- | --
647 | YT/RE

</figure></span></span>
</body>
</html>



What we expect to happen in those cases?

bigludo7:
Hello
If we could have a list of country for one MCC probably cleaner to have an array of countryName instead of one string and pipe to distinguish occurrence.
Thanks

eric-murray:
> If we could have a list of country for one MCC probably cleaner to have an array of countryName instead of one string and pipe to distinguish occurrence.

Also, for some MCC (principally 901), there is no associated country. So we need to decide whether a device roaming on an international network should return no countryName field at all, a blank countryName (`""`) or empty array (`[]`),  or some other value. I have seen `"XX"` used for "unknown" but this is a user-assigned code not part of the standard.

maxl2287:
> > If we could have a list of country for one MCC probably cleaner to have an array of countryName instead of one string and pipe to distinguish occurrence.
> 
> Also, for some MCC (principally 901), there is no associated country. So we need to decide whether a device roaming on an international network should return no countryName field at all, a blank countryName (`""`) or empty array (`[]`), or some other value. I have seen `"XX"` used for "unknown" but this is a user-assigned code not part of the standard.

Maybe for such 9xx we can go with "worldwide"?

ravindrapalaskar17:
> Hello If we could have a list of country for one MCC probably cleaner to have an array of countryName instead of one string and pipe to distinguish occurrence. Thanks

We updated data type of country name from string to array for multiple country name.

code/API_definitions/check-device-connectivity.yaml
```diff
@@ -158,6 +162,16 @@ components:
...
+    CountryCode:
+      description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
+      type: integer
+      example: 340
+    CountryName:
+      description: A list of ISO 3166 ALPHA-2 country-codes of mapped to the mobile country code (MCC).
+      type: array
+      items:
+        type: string
+      example: [BL,GF,GP,MF,MQ]
```
eric-murray, line 174: I'd like to see multiple examples here (using keyword `examples` rather than `example`):
- No associated country `[]`
- Single associated country e.g. `["DE"]`
- Multiple associated countries e.g. `["BL","GF","GP","MP","MQ"]`

The current example with multiple countries would be rare in practice. The usual response would be a single country, so the description should be extended to explain why multiple countries or no country may be returned.

My current preference for no associated country is an empty array, but an alternative would be `["XX"]`.

ravindrapalaskar17: Multiple Country Code
Single Country Code
No Country Code

ravindrapalaskar17: We Updated YAML with all three Examples with extended description.

monamok: Hello @ravindrapalaskar17 
I agree with @eric-murray about "No associated country" case. I think an empty array [] is a better choice. Is there any reason why you prefer to return ["XX"]?

code/API_definitions/check-device-connectivity.yaml
monamok: > we need to consider user consent.
> The user may consent to their roaming status being made available to the API caller, but not the country in which they are roaming. Providing the roaming country provides much more information on the device location than just knowing that the device is not in the home country.


You're right about user consent. How are we going to manage and seperate these cases? By having different scopes?

code/API_definitions/check-device-connectivity.yaml
maxl2287: @ravindrapalaskar17 let's go with an empty array for an undefined countryName.

ravindrapalaskar17: Thanks for feedback we updated YAML with empty array.

maxl2287 (approved):

code/API_definitions/check-device-connectivity.yaml
```diff
@@ -51,6 +51,34 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ConnectivityEventReport'
+              examples:
+                No-Country-Code:
+                  value:
+                    ueId:
+                      msisdn: 123456789
+                    uePort: 5090
```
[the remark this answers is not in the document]
ravindrapalaskar17: Update YAML:- Removed uePort from examples

eric-murray (approved):

eric-murray:
@NoelWirzius
Can this PR be merged?

NoelWirzius (approved):

NoelWirzius:
@eric-murray yes we can go merge this

eric-murray:
Now merged

Issue #22, opened 2023-02-15
Simplifying the API


monamok:
As we've been talking today during the meeting, there are issues in commonolities that they're still pending and as we want to move forward and there are already some new proposals for new requirements, it's better to decide if want to have a simpler version of the API, having dedicated endpoints and removing unnecessary parameters. These are the main topics:

Having smaller APIs vs general APIs that include many endpoints (all and everything is related to deviceStatus)
Dedicated endpoint vs general endpoints

We already have been talking about some of them before [here](https://github.com/camaraproject/DeviceStatus/pull/5#issuecomment-1294839976) but let’s have pros and cons and decide what is the best way to move forward. I opened a [PR](https://github.com/camaraproject/DeviceStatus/pull/23) because I find it easier commenting on each point but of course, we can continue the discussion here if you think it’s a better approach.

We have other 3 open issues:
-	[Align error message with structure defined in the guideline](https://github.com/camaraproject/DeviceStatus/issues/14)
-	[Update commonalities glossary with subproject relevant terms](https://github.com/camaraproject/DeviceStatus/issues/15)
-	[More Details for Roaming Status](https://github.com/camaraproject/DeviceStatus/issues/18)

So, in [this version](https://github.com/camaraproject/DeviceStatus/pull/23) I excluded the part we would change error definitions and rename the parameters according to the guidelines. Those issues will be included in separated PRs. And for the third one, there’s already an open PR and once it gets approved I’ll include it here as well.

I opened a new PR including the changes I’d like to suggest.
1.	Renaming the endpoint to have a dedicated endpoint for roaming status. 
2.	If the endpoint is dedicated to “roaming”, request body parameters eventType and eventStatus won’t be necessary anymore.
3.	Removing from response parameters that we already used to send the request to the endpoint. Is there any reason why we would need to get them back?
4.	The response was changed into a Boolean field to say if the device is roaming or not. We could keep the enum field with “ROAMING_ON" "ROAMING_OFF" values but do we really need it?
5.	The tag name and some descriptions were edited to be aligned with the changes that were made
6.	I didn’t change the filename. Depending on the result I’ll adjust it.

Please share your thoughts and preferences here or directly on the PR.

monamok:
Point 6 is still pending. I close this issue and open a new one to propose to fix this and a description typo that I found.

PR #23, opened 2023-02-15
Simplifying API

monamok:
This [issue](https://github.com/camaraproject/DeviceStatus/issues/22) was opened to keep track of this PR. Please review it and we can change it according to the comments and decisions that will be made.


monamok:
@NoelWirzius I don't have permissions to add you as a reviewer.

bigludo7 (approved):
Look good for me.

eric-murray:
Is the proposal that we have separate APIs (i.e. separate basePaths) per use case? So if connectivity status was re-introduced, that would be a new API with its own YAML definition?

monamok:
> Is the proposal that we have separate APIs (i.e. separate basePaths) per use case? So if connectivity status was re-introduced, that would be a new API with its own YAML definition?

Hello @eric-murray. Yes, the proposal is to have a portfolio of Device Status APIs. So we can call this API " Device Roaming Status" (which is the only use case that we currently have) and whenever we receive new requirements related to Device Connectivity Status we discuss them again and categorize them (if they’re not related to roaming) as new APIs with their own YAML files with dedicated endpoints. 
The users might not be interested in all use cases that can be grouped in a big API and if we gather different kind of status/connectivity in a single API they might be forced to implement the API partially. 
Does it work for you?

eric-murray:
@monamok
Thanks. There are advantages and disadvantages of both approaches. I just wanted to be sure I understood what was being proposed. I think I would prefer a common basePath for anything that was considered to be a "device status", even if each path (i.e. use case) was published in a separate YAML file. I'll discuss internally.

Another question is that PR #21 is also proposing changes to the API definition. Either could be merged at the moment, but after the first is merged, the other will need to be re-based. Is it agreed which should be merged first?

monamok:
> I'll discuss internally.
@eric-murray, of course. Please let us know.

@NoelWirzius can we please have also your feedback about this?

> Another question is that PR https://github.com/camaraproject/DeviceStatus/pull/21 is also proposing changes to the API definition. Either could be merged at the moment, but after the first is merged, the other will need to be re-based. Is it agreed which should be merged first?

I already mentioned this point in the issue [#22] (https://github.com/camaraproject/DeviceStatus/issues/22) that I opened for this PR. If PR [#21](https://github.com/camaraproject/DeviceStatus/pull/21) get merged before this one, I'll update my branch with the approved changes so there's no conflict. The only thing we should decide is if we keep the version or should we update it too?

eric-murray:
@monamok
I discussed with our developers, and they much prefer to have separate basepaths for different API definitions (i.e. different YAML files). So I'm happy to tie the basepath to the use case (i.e. `/device-roaming-status`). though do wonder if we should keep the path itself as `status` rather than also changing that to `roaming`.

I'm also awaiting feedback from @NoelWirzius on the order in which PRs should be treated before approving this one.

NoelWirzius:
@eric-murray @monamok 

we checked the different pro and cons and the cons are really big in our views. So the maintenance of the APIs will get really complex when we build for everything an own API - then we need to do changes to each API for example changing the naming or adding a new function like "list of devices"

but there are also pros of course

Our suggestion is:
- stay at one YAML and adding different Endpoints for each API

Example for Roaming, Connectivity and subscription:

server/devicestatus/v1/roaming
server/devicestatus/v1/connectivity 
server/devicestatus/v1/subscriptions

is this sth we can go as a compromise?

NoelWirzius:
@eric-murray  fully agree with you! the same feedback I also got. If we split -> then different YAMLs - but we would prefer a single common YAML with a common basepath and just different endpoints for each "Feature"

bigludo7:
I'm also in favor of a common YAML in order to be consistent with other API with several 'sub' function as we have this pattern in number verification API and SIM Swap API. The YAML covers the 'family' and we have several endpoints.

We have distinct YAML for Carrier billing/SIM SWAP/Number Verification when we have distinct flavor of the requirement  (MC vs specific, oma vs specific) but same functions are covered in each YAML.

monamok:
Thank you @eric-murray @NoelWirzius @bigludo7 for your feedback. Sorry I was trying to resolve the conflicts and I accidentally closed the PR. I'm not sure if someone from the code owners can reopen it or should I open a new PR? 
I aligned the YAML file with your comments and I already pushed it to my repo in order to have a dedicated endpoint for roaming as @NoelWirzius suggested with a common base path for future usecases.

monamok:
> I'm not sure if someone from the code owners can reopen it or should I open a new PR?
I was able myself to reopen it :)

sachinvodafone:
Hi Team , if we will go with single YAML file for all related cases (device status) then how the basepath will be defined to create proxy. It would be challenging to have different proxy for individual endpoint (Roaming, Connectivity etc)  and with single porxy, it will make our job more tough to manage Spike, authentication/Authorisation and other security policies for specific requirement.

eric-murray:
I'm approving this because it keeps open the question of whether additional statuses (such as connectivity) should be in the same or a separate YAML. When an addition status is proposed, that issue can be addressed then.

eric-murray (approved):

bigludo7 (approved):
This is good for me.

monamok:
@MarkusKuemmerle could you please add again @NoelWirzius as a reviewer. I think he was removed automatically when the PR was reopened. Thanks.

PR #25, opened 2023-02-28
Multiple devices roaming status

KaKi458:
Multiple devices can be requested in one call -  roaming status is returned for each one


eric-murray:
@KaKi458 Please raise this as an issue for discussion first

Issue #17, opened 2023-01-31
Subscription Mode with new Eventtypes 


NoelWirzius:
**Summary**
===========
In this issue will add a new **path** which enables subscription mode. On top of this is the enablement of new events and the multiple handling of devices and events. 

Due to this it is possible to get more informations about a device based on a subscription level. There are several Use-Cases like Streaming and Fraud where a "push" from the network can trigger events on the application sides

**Changes** 
==========
* Adding new Path `/Subscriptions`
  *  Eventtypes
    * `LOSS_OF_CONNECTIVITY`
      * This will give back if the device is loosing the connection to the network (SMS/DATA)
    * `UE_REACHABILITY`
      * This will give back when a device is gaining connection to the network 
    *  `ROAMING_STATUS` 
      * Informs when roaming status changed  
    * Attributes
      * `expireTime` 
        * Defines how long the subscription should run
    * Multiple Devices 
      * Enable to set subscription directly for a complete list of identifiers 
    * Multiple Eventtypes
      *  Enable to set subscription for identifiers to multiple events

bigludo7:
Hello @NoelWirzius 
I'm interested to contribute for this issue to improve the current API with adding notification support.

About the event, I've raised an issue in commonalities to unify the event structure (https://github.com/camaraproject/WorkingGroups/issues/156) and I'm looking for feedback.
Thanks

eric-murray:
I disagree with the proposed approach in PR #28. There is already a `POST \roaming` path defined, so why not allow the client to specify whether they want an immediate response or to set-up an ongoing subscription? This could be determined explicitly (e.g. via a flag) or implicitly (e.g. given the presence or absence of subscription parameters). Immediate response returns 200 and subscription returns 201.

This also keeps open the option to have additional statuses in different YAMLs.

bigludo7:
Hello @eric-murray 

Thanks for your review!

For me this is distinct requirement - and distinct services from Telco monetization perspective - between managing one-time request/response to an event-base request that could stay 'alive' for a long time (based on the subscriptionExpireTime). 

We are working with this approach in the device location API where you can query a device location now or subscribe to get notification when the device entered/left an area.

I'm not sure to get your point about : _This also keeps open the option to have additional statuses in different YAMLs_.  you mean managing `loss of connectivity`  and `reachability` in another swagger?

Let's continue the discussion and waiting for other feedback - probably we can tab a discussion for our next meeting (March 15th )

[...]

PR #31, opened 2023-03-13
Adding notification/subscription for Device Status

bigludo7:
New proposal following our discussion on issue  #17 . 

Not for merging at this stage but to get team feedback - I will like to use the work we're doing here to contribute to API guidelines for subscription/notification model.

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -32,21 +32,22 @@ servers:
 tags:
   - name: Device roaming status
     description: Operation to get device roaming status and country information (if roaming) synchronously
-  - name: Device status subscription
-    description: Operation to manage subscription on device status event (roaming, connectivity)
+  - name: Device status event subscription
+    description: Operation to manage event subscription on device status event (roaming, connectivity)
   - name: Notification Callback
     description: |
       Notification received on subscription listener side. 
       
       **WARNING**: This endpoint must be exposed on the listener side as POST /notifications 
-      will be posted by the Roaming resource server to the url provided by notification 
+      will be posted by the device status resource server to the url provided by notification 
       listener.
 paths:
   /notifications:
     post:
       tags:
         - Notification Callback
-      summary: "Subscription notification callback"
+      summary: "Notification endpoint to notify listener that a subscribed event occurred"
+      description: Notification endpoint to notify listener that a subscribed event occurred. This endpoint must be exposed on the listener side.
       operationId: postNotification
       requestBody:
         required: true
@@ -63,8 +64,10 @@ paths:
                 $ref: '#/components/examples/ROAMING_ON'
               ROAMING_OFF:
                 $ref: '#/components/examples/ROAMING_OFF'
+              SUBSCRIPTION_ENDS:
+                $ref: '#/components/examples/SUBSCRIPTION_ENDS'
       responses:
-        '204':
+        204:
           description: No Content
         400:
           $ref: "#/components/responses/Generic400"
@@ -85,6 +88,7 @@ paths:
       tags:
         - Device roaming status
       summary: 'Get the current roaming status and the country information'
+      description: Get the current roaming status and the country information
       operationId: getRoamingStatus
       requestBody:
         content:
@@ -93,7 +97,7 @@ paths:
               $ref: '#/components/schemas/RequestRoamingStatus'
         required: true
       responses:
-        '200':
+        200:
           description: Contains information about current roaming status
           content:
             application/json:
@@ -115,84 +119,92 @@ paths:
                     roaming: true
                     countryCode: 340
                     countryName: [ "BL","GF","GP","MF","MQ" ]
-        '400':
+        400:
           $ref: "#/components/responses/Generic400"
-        '401':
+        401:
           $ref: "#/components/responses/Generic401"
-        '403':
+        403:
           $ref: "#/components/responses/Generic403"
-        '404':
+        404:
           $ref: "#/components/responses/Generic404"
-        '500':
+        500:
           $ref: "#/components/responses/Generic500"
-        '503':
+        503:
           $ref: "#/components/responses/Generic503"
   /event-subscriptions:
     post:
       tags:
-        - Device status subscription
-      summary: 'Create a device status subscription for a device'
-      operationId: createDevicStatusSubscription
+        - Device status event subscription
+      summary: 'Create a device status event subscription for a device'
+      description: Create a device status event subscription for a device
+      operationId: createDevicStatusEventSubscription
       requestBody:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/CreateSubscription'
+              $ref: '#/components/schemas/CreateEventSubscription'
         required: true
       responses:
-        '201':
+        201:
           description: Created
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/SubscriptionInfo'
-        '400':
+                $ref: '#/components/schemas/EventSubscriptionInfo'
+        202 :
+          description: Request accepted to be processed. It applies for async creation process.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EventSubscriptionAsync'
+        400:
           $ref: '#/components/responses/Generic400'
-        '401':
+        401:
           $ref: '#/components/responses/Generic401'
-        '403':
+        403:
           $ref: '#/components/responses/Generic403'
-        '409':
+        409:
           $ref: '#/components/responses/Generic409'
-        '500':
+        500:
           $ref: '#/components/responses/Generic500'
-        '503':
+        503:
           $ref: '#/components/responses/Generic503'
     get:
       tags:
-        - Device status subscription
-      summary: 'Retrieve a list of device status subscription'
-      operationId: retrieveSubscriptionList
+        - Device status event subscription
+      summary: 'Retrieve a list of device status event subscription'
+      description: Retrieve a list of device status event subscription(s)
+      operationId: retrieveEventSubscriptionList
       responses:
-        '200':
-          description: List of subscription details
+        200:
+          description: List of event subscription details
           content:
             application/json:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/SubscriptionInfo'
-        '400':
+                  $ref: '#/components/schemas/EventSubscriptionInfo'
+        400:
           $ref: '#/components/responses/Generic400'
-        '401':
+        401:
           $ref: '#/components/responses/Generic401'
-        '403':
+        403:
           $ref: '#/components/responses/Generic403'
-        '500':
+        500:
           $ref: '#/components/responses/Generic500'
-        '503':
+        503:
           $ref: '#/components/responses/Generic503'
-  /event-subscriptions/{subscriptionId}:
+  /event-subscriptions/{eventSubscriptionId}:
     get:
       tags:
-        - Device status subscription
-      summary: 'Retrieve a device status subscription for a device'
-      operationId: retrieveSubscription
-      description: retrieve subscription information for a given subscription.
+        - Device status event subscription
+      summary: 'Retrieve a device status event subscription for a device'
+      operationId: retrieveEventSubscription
+      description: retrieve event subscription information for a given subscription.
       parameters:
-        - name: subscriptionId
+        - name: eventSubscriptionId
           in: path
-          description: subscription identifier that was obtained from the create subscription operation
+          description: Event subscription identifier that was obtained from the create subscription operation
           required: true
           schema:
             type: string
@@ -202,7 +214,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/SubscriptionInfo'
+                $ref: '#/components/schemas/EventSubscriptionInfo'
         400:
           description: Invalid input
           content:
@@ -217,11 +229,11 @@ paths:
                     status: 400
                     message: "Schema validation failed at  ..."
                 subscriptionIdRequired:
-                  summary: subscription id is required
+                  summary: eventSubscription id is required
                   value:
                     code: INVALID_INPUT
                     status: 400
-                    message: "Expected property is missing: subscriptionId"
+                    message: "Expected property is missing: eventSubscriptionId"
         401:
           $ref: "#/components/responses/Generic401"
         403:
@@ -238,20 +250,26 @@ paths:
           $ref: "#/components/responses/Generic503"
     delete:
       tags:
-        - Device status subscription
-      summary: 'Delete a device status subscription for a device'
-      operationId: deleteSubscription
-      description: delete a  given subscription.
+        - Device status event subscription
+      summary: 'Delete a device status event subscription for a device'
+      operationId: deleteEventSubscription
+      description: delete a  given event subscription.
       parameters:
-        - name: subscriptionId
+        - name: eventSubscriptionId
           in: path
-          description: subscription identifier that was obtained from the create subscription operation
+          description: Event subscription identifier that was obtained from the create event subscription operation
           required: true
           schema:
             type: string
       responses:
-        200:
-          description: subscription deleted
+        204:
+          description: event subscription deleted
+        202 :
+          description: Request accepted to be processed. It applies for async deletion process.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EventSubscriptionAsync'
         400:
           description: Invalid input
           content:
@@ -266,11 +284,11 @@ paths:
                     status: 400
                     message: "Schema validation failed at  ..."
                 subscriptionIdRequired:
-                  summary: subscription id is required
+                  summary: eventSubscription id is required
                   value:
                     code: INVALID_INPUT
                     status: 400
-                    message: "Expected property is missing: subscriptionId"
+                    message: "Expected property is missing: eventSubscriptionId"
         401:
           $ref: "#/components/responses/Generic401"
         403:
@@ -424,12 +442,14 @@ components:
           $ref: '#/components/schemas/Device'
       required:
         - device
+
     Port:
       description: User equipment port. Device port may be required along with IP address to identify the target device
       type: integer
       minimum: 0
       maximum: 65535
       example: 5060
+
     ErrorInfo:
       type: object
       required:
@@ -446,29 +466,15 @@ components:
         message:
           type: string
           description: Detailed error description
-    DeviceStatusEventType:
-      description: Type of the event
-      type: string
-      enum:
-        - ROAMING_STATUS
-        - ROAMING_ON
-        - ROAMING_OFF
-        - ROAMING_CHANGE_COUNTRY
-    DeviceStatusEventTypeList:
-      description: List of Device Status Event Type
-      type: array
-      items:
-        $ref: '#/components/schemas/DeviceStatusEventType'
-      example: [ROAMING_STATUS]
-
-    CreateSubscription:
-      description: The request for creating a Device Status Subscription
+        
+    CreateEventSubscription:
+      description: The request for creating a Device Status event subscription
       type: object
       required:
         - webhook
       properties:
         subscriptionDetail:
-          $ref: '#/components/schemas/SubscriptionDetail'
+          $ref: '#/components/schemas/EventSubscriptionDetail'
         subscriptionExpireTime:
           type: string
           format: date-time
@@ -486,14 +492,15 @@ components:
         notificationUrl:
           type: string
           example: "https://application-server.com"
-          description: Allows asynchronous delivery of device status related events
+          description: https callback address where the event notification must be POST-ed
         notificationAuthToken:
           type: string
           example: "c8974e592c2fa383d4a3960714"
-          description: Authentification token for callback API
+          description: | 
+            OAuth2 token to be used by the callback API endpoint. It MUST be indicated within HTTP Authorization header e.g. Authorization: Bearer $notificationAuthToken
 
-    SubscriptionDetail:
-      description: The detail of the requested description
+    EventSubscriptionDetail:
+      description: The detail of the requested event subscription
       type: object
       required:
         - eventType
@@ -506,27 +513,27 @@ components:
     RoamingEventType:
       type: string
       description: |
-        ROAMING_STATUS - Device status event triggered when the UE switch from roaming ON to roaming OFF and conversely
+        ROAMING_STATUS - Event triggered when the device switch from roaming ON to roaming OFF and conversely
 
-        ROAMING_ON - Device status event triggered when the UE switch from roaming OFF to roaming on
+        ROAMING_ON - Event triggered when the device switch from roaming OFF to roaming ON
 
-        ROAMING_OFF: Device status event triggered when the UE switch from roaming ON to roaming OFF
+        ROAMING_OFF: Event triggered when the device switch from roaming ON to roaming OFF
 
-        ROAMING_CHANGE_COUNTRY: Device status event triggered when the UE in roaming change country code|
+        ROAMING_CHANGE_COUNTRY: Event triggered when the device in roaming change country code
       enum:
         - ROAMING_STATUS
         - ROAMING_ON
         - ROAMING_OFF
         - ROAMING_CHANGE_COUNTRY
 
-    SubscriptionInfo:
+    EventSubscriptionInfo:
       description: Represents a device status subscription.
       allOf:
-        - $ref: '#/components/schemas/CreateSubscription'
+        - $ref: '#/components/schemas/CreateEventSubscription'
         - type: object
           properties:
-            subscriptionId:
-              $ref: '#/components/schemas/SubscriptionId'
+            eventSubscriptionId:
+              $ref: '#/components/schemas/EventSubscriptionId'
             startsAt:
               type: string
               format: date-time
@@ -538,21 +545,32 @@ components:
           required:
             - subscriptionId
             - type
-    SubscriptionId:
+
+    EventSubscriptionAsync:
+      description: Response for a device status operation managed asynchronously (Creation or Deletion)
+      type: object
+      properties:
+          eventSubscriptionId:
+              $ref: '#/components/schemas/EventSubscriptionId'
+
+    
+    EventSubscriptionId:
       type: string
-      description: The subscription identifier.
+      description: The event subscription identifier.
+    
     EventTime:
       format: date-time
       type: string
       description: The time when the event was notified.
+    
     Notification:
       description: The notification callback.
       type: object
       required:
         - event
       properties:
-        subscriptionId:
-          $ref: '#/components/schemas/SubscriptionId'
+        eventSubscriptionId:
+          $ref: '#/components/schemas/EventSubscriptionId'
         event:
           $ref: '#/components/schemas/Event'
 
@@ -579,6 +597,7 @@ components:
           ROAMING_CHANGE_COUNTRY : '#/components/schemas/EventChangeCountry'
           ROAMING_OFF: '#/components/schemas/EventRoamingOff'
           ROAMING_ON: '#/components/schemas/EventRoamingOn'
+          SUBSCRIPTION_ENDS : '#/components/schemas/EventSubscriptionEnds'
 
     EventRoamingStatus:
       description:  event structure for roaming status change
@@ -628,6 +647,15 @@ components:
             eventDetail:
               $ref: '#/components/schemas/RoamingChangeCountry'
 
+    EventSubscriptionEnds:
+      description:  event structure for event subscription ends
+      allOf:
+        - $ref: '#/components/schemas/Event'
+        - type: object
+          properties:
+            eventDetail:
+              $ref: '#/components/schemas/SubscriptionEnds'
+
     RoamingStatus:
       description: Event detail structure for ROAMING_STATUS event
       type : object
@@ -666,6 +694,15 @@ components:
           $ref: '#/components/schemas/CountryCode'
         countryName:
           $ref: '#/components/schemas/CountryName'
+
+    SubscriptionEnds:
+      description: Event detail structure for SUBSCRIPTION_ENDS event
+      type : object
+      required:
+        - device
+      properties:
+        device:
+          $ref: '#/components/schemas/Device'
 
   responses:
     Generic400:
@@ -778,3 +815,12 @@ components:
               phoneNumber: 123456789
             countryCode: 214
             countryName: SP
+    SUBSCRIPTION_ENDS:
+      value:
+        subscriptionId: 123655
+        event:
+          eventType: SUBSCRIPTION_ENDS
+          eventTime: 2023-01-17T13:18:23.682Z
+          eventDetail:
+            device:
+              phoneNumber: 123456789
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -31,13 +31,16 @@ servers:
         description: Base path for the device status API
 tags:
   - name: Device roaming status
-    description: Operations to get device roaming status and country information on the fly
-  - name: Device roaming status subscription
-    description: Operation to manage subscription on device status roaming event
+    description: Operation to get device roaming status and country information (if roaming) synchronously
   - name: Device status subscription
     description: Operation to manage subscription on device status event (roaming, connectivity)
   - name: Notification Callback
-    description: Notification received on subscription listener side
+    description: |
+      Notification received on subscription listener side. 
+      
+      **WARNING**: This endpoint must be exposed on the listener side as POST /notifications 
+      will be posted by the Roaming resource server to the url provided by notification 
+      listener.
 paths:
   /notifications:
     post:
@@ -63,6 +66,20 @@ paths:
       responses:
         '204':
           description: No Content
+        400:
+          $ref: "#/components/responses/Generic400"
+        401:
+          $ref: "#/components/responses/Generic401"
+        403:
+          $ref: "#/components/responses/Generic403"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
+          $ref: "#/components/responses/Generic503"
   /roaming:
     post:
       tags:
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Fixed small issue

bigludo7 (commented):
Removed mention to LOSS OF CONNECTIVITY & REACHABILITY
Fixed example.

code/API_definitions/device-status.yaml
```diff
@@ -168,7 +354,7 @@ components:
       description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
       type: integer
     CountryName:
-      description: The ISO 3166 ALPHA-2 country-codes of mapped to mobile country code(MCC). If there is mapping of one MCC to multiple countries, then we have list of countries. If there is no mapping of MCC to any country, then an empty array [] shall be returned.
+      description: The ISO 3166 ALPHA-2 country-codes of mapped to mobile country code(MCC). If there is mapping of one MCC to multiple countries, then we have list of countries. If there is no mapping of MCC to any country, then ["xx"] will be used.
```
monamok, line 357: @bigludo7 I forgot to mention this in the meeting. Here it's overwriting the last change that was added to the main branch. Could you please align it?

bigludo7 (commented):
Updated:
- Align with Subscription &  Notification model
- Splitted in 4 event type for roaming

bigludo7 (commented):
Solve issue https://github.com/camaraproject/DeviceStatus/issues/52

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +398,205 @@ components:
...
+          $ref: '#/components/schemas/SubscriptionDetail'
+        subscriptionExpireTime:
+          type: string
+          format: date-time
+          example: 2023-01-17T13:18:23.682Z
+          description: The subscription expiration time in date-time format.
+        webhook:
+          $ref: '#/components/schemas/Webhook'
+    Webhook:
+      description: Webhook information for event channel
```
monamok, line 429: According to the guidlines it's mandatory.

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +398,205 @@ components:
...
+        - ROAMING_STATUS
+        - ROAMING_ON
+        - ROAMING_OFF
+        - ROAMING_CHANGE_COUNTRY
+    DeviceStatusEventTypeList:
+      description: List of Device Status Event Type
+      type: array
+      items:
+        $ref: '#/components/schemas/DeviceStatusEventType'
+      example: [ ROAMING_STATUS]
```
monamok, line 414:
```suggestion
      example: [ROAMING_STATUS]
```

code/API_definitions/device-status.yaml
```diff
@@ -30,9 +31,39 @@ servers:
...
+            examples:
+              ROAMING_STATUS:
+                $ref: '#/components/examples/ROAMING_STATUS'  
+              ROAMING_CHANGE_COUNTRY:
+                $ref: '#/components/examples/ROAMING_CHANGE_COUNTRY' 
+              ROAMING_ON:
+                $ref: '#/components/examples/ROAMING_ON'
+              ROAMING_OFF:
+                $ref: '#/components/examples/ROAMING_OFF'
+      responses:
```
monamok, line 63: From the guideline:

_Error definition are described in this guideline applies for event notification.
Following Error code must be present:_

    for POST: 400, 401, 403, 500, 503

monamok (commented):
Thank you @bigludo7.
I added some comments comparing it with the guidlines and some additional point to discuss with the working group.

In general I suggest we use parameters descriptions from Glossary with some adjustments but they might be helpful for the developers.

code/API_definitions/device-status.yaml
```diff
@@ -120,38 +310,43 @@ components:
...
-      format: ipv4
-      pattern: '^([01]?\d\d?|2[0-4]\d|25[0-5])(?:\.(?:[01]?\d\d?|2[0-4]\d|25[0-5])){3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$'
-      example: '192.168.0.1/24'
-      description: |
-        IPv4 address may be specified in form <address/mask> as:
-          - address - an IPv4 number in dotted-quad form 1.2.3.4. Only this exact IP number will match the flow control rule.
-          - address/mask - an IP number as above with a mask width of the form 1.2.3.4/24.
-            In this case, all IP numbers from 1.2.3.0 to 1.2.3.255 will match. The bit width MUST be valid for the IP version.
-    Ipv6Addr:
+    Ipv4Address:
```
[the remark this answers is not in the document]
eric-murray: Allowing `privateAddress` as a device identifier makes it easier for the device itself (or an app on the device) to make the API call. But for this API, I don't see why the device itself would need to use an API call to the network to determine its roaming (or connectivity) status. So I don't think it is needed here to support that scenario.

Having said that, an API consumer other than the device itself may know the private IP address, and so this would still be a valid way of identifying the device. In fact, it is easier for the network to identify the device by `privateAddress` + `publicAddress` compared to `publicAddress` + `publicPort`, so we should encourage use of `privateAddress` if that is known.

So my preference is to follow QoD and Device Location when it comes to identifying the device by `privateAddress`, even if the justification is not so strong.

code/API_definitions/device-status.yaml
```diff
@@ -79,8 +110,167 @@ paths:
...
+      summary: 'Create a device roaming status subscription for a device'
+      operationId: createRoamingSubscription
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateRoamingSubscription'
+        required: true
+      responses:
+        '201':
```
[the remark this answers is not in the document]
sachinvodafone: I think that event subscription should be performed synchronously, and if the server times out while waiting for the request to finish, we can respond with a 408 error code.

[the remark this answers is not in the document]
bigludo7: I'm on same page than @sachinvodafone that the subscription must be sync. For the error I propose to stick to the guidelines (I will make sure to have all) 
for POST: 400, 401, 403, 409, 500, 503
for GET: 400, 401, 403, 500, 503
for GET/{subscriptionId}: 400, 401, 403, 404, 500, 503
for DELETE: 400, 401, 403, 404, 500, 503

code/API_definitions/device-status.yaml
```diff
@@ -79,8 +110,167 @@ paths:
...
+      description: delete a  given subscription.
+      parameters:
+        - name: subscriptionId
+          in: path
+          description: subscription identifier that was obtained from the create subscription operation
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
```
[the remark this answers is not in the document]
bigludo7: For me a `DELETE `operation on a subscription must be done synchronously. We did not define a specifically async mechanism in the guideline for that `DELETE `operation so I presume to keep it sync.

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
monamok: But we could allow both, couldn't we? Why do you think we should force them to be synchronous and not allow to choose?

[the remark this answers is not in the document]
bigludo7: @monamok  But why you want this for this particular operation? 
The event subscription is a standard request not related by asynchronous mechanism itself. If we add this, we should also add to POST / roaming  and all other POST operation.
The only exception for me is the one for the notification itself.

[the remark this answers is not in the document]
sachinvodafone: While it is possible to implement asynchronous operations for both GET and POST operations across all endpoints in the Camara API, it's important to note that certain operations may inherently require immediate responses. These operations are better suited for synchronous handling. It is necessary for us to align and agree on this whether it is best suited for Syn or Asy , as this consideration will apply to every use case.

bigludo7 (commented):
Updated the proposal after @monamok review (huge thanks for your detailed review 🥇 ) , integrated outcomes of meeting discussion to simplify the path of the subscription, and as suggested by @eric-murray aligned device definition with QoD API.

code/API_definitions/device-status.yaml
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -30,9 +31,39 @@ servers:
...
 tags:
   - name: Device roaming status
-    description: Operations to validate device roaming status and country information
+    description: Operations to get device roaming status and country information on the fly
+  - name: Device roaming status subscription
+    description: Operation to manage subscription on device status roaming event
+  - name: Device status subscription
+    description: Operation to manage subscription on device status event (roaming, connectivity)
+  - name: Notification Callback
+    description: Notification received on subscription listener side
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -170,19 +365,17 @@ components:
       description: The Mobile country code (MCC) as an geographic region identifier for the country and the dependent areas.
       type: integer
     CountryName:
-      description: The ISO 3166 ALPHA-2 country-codes of mapped to mobile country code(MCC). If there is mapping of one MCC to multiple countries, then we have list of countries. If there is no mapping of MCC to any country, then an empty array [] shall be returned.
+      description: The ISO 3166 ALPHA-2 country-codes of mapped to mobile country code(MCC). If there is mapping of one MCC to multiple countries, then we have list of countries. If there is no mapping of MCC to any country, then ["xx"] will be used.
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +398,205 @@ components:
...
+          description: Authentification token for callback API  
+    SubscriptionDetail:
+      description: The detail of the requested description
+      type: object
+      properties:
+        device:
+          $ref: '#/components/schemas/Device'
+        eventType:
+          $ref: '#/components/schemas/RoamingEventType'
+    RoamingEventType:
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -79,8 +110,167 @@ paths:
           $ref: "#/components/responses/Generic500"
         '503':
           $ref: "#/components/responses/Generic503"
-        '504':
-          $ref: "#/components/responses/Generic504"
+  /roaming/event-subscription:
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
bigludo7: updated

code/API_definitions/device-status.yaml
```diff
@@ -120,38 +310,43 @@ components:
...
+          $ref: '#/components/schemas/NetworkAccessIdentifier'
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        ipv4Address:
+          $ref: '#/components/schemas/Ipv4Address'
+        ipv6Address:
+          $ref: '#/components/schemas/Ipv6Address'
       minProperties: 1
-    ExternalId:
+    NetworkAccessIdentifier:
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +398,205 @@ components:
...
+          format: date-time
+          example: 2023-01-17T13:18:23.682Z
+          description: The subscription expiration time in date-time format.
+        webhook:
+          $ref: '#/components/schemas/Webhook'
+    Webhook:
+      description: Webhook information for event channel
+      type: object
+      properties: 
+        notificationUrl:
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -79,8 +110,167 @@ paths:
...
+                type: array 
+                items:
+                  $ref: '#/components/schemas/SubscriptionInfo'
+        '400':
+          $ref: '#/components/responses/Generic400'
+        '401':
+          $ref: '#/components/responses/Generic401'
+        '403':
+          $ref: '#/components/responses/Generic403'
+        '409':
```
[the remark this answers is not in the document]
bigludo7: Fixed

code/API_definitions/device-status.yaml
bigludo7: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -30,9 +31,56 @@ servers:
...
+  - name: Notification Callback
+    description: |
+      Notification received on subscription listener side. 
+      
+      **WARNING**: This endpoint must be exposed on the listener side as POST /notifications 
+      will be posted by the Roaming resource server to the url provided by notification 
+      listener.
 paths:
-   /roaming:
+  /notifications:
```
akoshunyadi, line 45: shouldn't we put the notification endpoint inside POST /event-notifications as callback?

bigludo7: @akoshunyadi - This is a great question. We saw that QoD API went in this direction;

My developer colleagues are not comfortable with this pattern (have the callback in the end point) as it is not really supported by code generator. A good point to discuss in Commonalities?

bigludo7:
Updated swagger (event model) after @patrice-conil code review.

code/API_definitions/device-status.yaml
```diff
@@ -117,47 +321,78 @@ components:
...
       properties:
-        externalId:
-          $ref: '#/components/schemas/ExternalId'
-        msisdn:
-          $ref: '#/components/schemas/MSISDN'
-        ipv4Addr:
-          $ref: '#/components/schemas/Ipv4Addr'
-        ipv6Addr:
-          $ref: '#/components/schemas/Ipv6Addr'
+        phoneNumber:
```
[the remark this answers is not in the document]
eric-murray: CAMARA does not use IMSI as a mobile subscription identifier for any API

code/API_definitions/device-status.yaml
```diff
@@ -79,8 +127,164 @@ paths:
...
+          $ref: '#/components/responses/Generic400'
+        '401':
+          $ref: '#/components/responses/Generic401'
+        '403':
+          $ref: '#/components/responses/Generic403'
+        '500':
+          $ref: '#/components/responses/Generic500'
+        '503':
+          $ref: '#/components/responses/Generic503'
+  /event-subscriptions/{subscriptionId}:
```
monamok, line 185: eventSubscriptionId?

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
monamok: I understand the point of view of having synchronous response in create/deleteSubscriptions. 
Having said that, during notification guidelines track, as per PR https://github.com/camaraproject/WorkingGroups/pull/173, the response codes for successful operations were not discussed. Thus no decision was made about including only synchronous operations or we could include both mechanisms for “write” subscription operations (in case of this API. POST -createSubscription-, DELETE -deleteSubscription-). In subscription section of the guideline specifies the error codes for eachpoint, but not the success codes. There are some examples given for success cases, but other than that, no clear instruction which says we can't consider asynchronous responses.

On the other hand, in https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#32-http-response-codes, it is reflected the use of 202 Accepted for asynchronous process.

In Telefonica, an asynchronous-based subscription model is designed, because this “write” operations manage actions in several underlying systems and that could take some time. For that reason, we would like to support both:

- 201 and 202 for create (we are returning required parameter eventSubscriptionId anyway)
- 202 and 204 for delete 

In case we allow 202 responses, we should mention in the guideline that GET /subscriptions/{subscriptionId} must be used to ensure existence of the subscription.

monamok (commented):
Thank you for all the effort @bigludo7. 

I added some more comments.

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +454,210 @@ components:
...
+            subscriptionId:
+              $ref: '#/components/schemas/SubscriptionId'
+            startsAt:
+              type: string
+              format: date-time
+              description: date time when subscription started
+            expiresAt:
+              type: string
+              format: date-time
+              description: date time when subscription will expire or expired
```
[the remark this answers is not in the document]
bigludo7: Not sure to understand the comment. `subscriptionExpireTime` is the expiration date wished by the requester. Suppose you want to have one month subscription. 
for GET I'm not sure we should enforce  this rule. As subscription will be monetized should we keep 'for some time' before to delete them?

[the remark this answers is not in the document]
bigludo7: From our discussion, I change `expiresAt `to `date time when subscription expired`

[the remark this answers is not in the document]
monamok: No, I think it should be "date time when subscription will expire". We are creating a subscriptionand the response says, this subscription starts at "x time" will expire in "y time", aren't we?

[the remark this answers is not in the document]
bigludo7: Not sure we're aligned....

The requester can provide a date wished for subscription: `subscriptionExpireTime `; fine for that.

July 5th, 2023 requester asks for a subscription till July 25,2023  - in the subscription both `subscriptionExpireTime  `& `expireDate `will be valued to July 25th.

Subscription server could also fill a future expire date. For example a subscription server did no support subscription for more the 3 months. If today is July 5th, & requester did no provided `subscriptionExpireTime  `or `subscriptionExpireTime`>3 month from now, then the `expireDate `will be October 5th. 

I assume at this point we're aligned :)

Now, what are we expecting once the `expireDate ` is reached?
Do we assume that the subscription resource is deleted and not anymore accessible by `GET`? or does this resource is still accessible for Tracking for example? I'm not sure about an unified pattern and that's why at last I prefer the initial wording: Date time when subscription expired or will expire.

bigludo7 (commented):
Update swagger after @monamok review

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +446,227 @@ components:
...
+    Webhook:
+      description: Webhook information for event channel
+      type: object
+      required:
+        - notificationUrl
+      properties:
+        notificationUrl:
+          type: string
+          example: "https://application-server.com"
+          description: Allows asynchronous delivery of device status related events
```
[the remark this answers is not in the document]
bigludo7: done.

sachinvodafone (approved):

code/API_definitions/device-status.yaml
eric-murray: My preference is to have the notification endpoint defined as a callback. However, I understand that code generators can have issues with callbacks, and having the separate notifications path does comply with the design guidelines (though these make it look like the endpoint should be called `eventNotifications` - maybe some editing is needed there). So I will approve this PR so that we can move towards v0.5.0.

Ongoing, I agree this would be a good topic to raise in Commonalities. I have seen three methods for defining notifications discussed:
- As a callback (as we have in QoD API)
- As a separate notifications path, with appropriate documentation (as we have here)
- As a completely separate OAS definition for use by the API consumer notification endpoint

I don't think these methods are incompatible. Why not have all three? But this may also cause issues with code generators if both a callback and separate notification path are defined, so should be discussed in Commonalities.

eric-murray (approved):

code/API_definitions/device-status.yaml
patrice-conil: Our proposal with @bigludo7 is to provide the definition as a callback because it documents very well what is  happening (and will align with how it's done in QoD). Also add a separate OAS definition to help implementers of this callback.
As you say, these approaches are not incompatible, so we can benefit from them.
We agree with you to discuss this point in the commonalities to help harmonization between APIs.

eric-murray: Thanks @patrice-conil

So is the proposal to modify this PR to include the notification endpoint as a callback? Or will that wait until after the discussion in Commonalities?

patrice-conil: @eric-murray,
I think it can wait until after discussion in Commonalities. I will propose a new PR after discussion if necessary.

eric-murray: Thanks @patrice-conil

Once there are the additional approvals from @monamok and @akoshunyadi, I can merge this.

code/API_definitions/device-status.yaml
```diff
@@ -389,71 +466,75 @@ components:
...
-    SubscriptionInfo:
+        device:
+          $ref: '#/components/schemas/Device'
+        eventType:
+          $ref: '#/components/schemas/RoamingEventType'
+
+    RoamingEventType:
+      type: string
+      description: |
+        ROAMING_STATUS - Event triggered when the device switch from roaming ON to roaming OFF and conversely
```
JoachimDahlgren, line 516: Looking at the event generated by a ROAMING_STATUS event subscription, both ON/OFF as well as country code/name is returned. 

I thereby assume that subscribing to ROAMING_STATUS will give you notifications for on, off as well change country. 

For clarity this should be reflected in the description.

eric-murray: If I, with a UK SIM, were in, for example, France and travelled to Germany, my `roaming` state would remain unchanged at `true`. So I would not expect a `ROAMING_STATUS` change event to be generated.

But if this is not clear, then indeed the description should be improved.

JoachimDahlgren: Good comment Eric. Then I am wondering about the ROAMING_CHANGE_COUNTRY event type. That clearly indicate that I would get a roaming status notification in the situation you describe above to indicate that you move from one country to another while roaming. 

I was thinking of the ROAMING_ON, ROAMING_OFF and ROAMING_CHANGE_COUNTRY as a subset of ROAMING_STATUS. Maybe that is not the case after all?

code/API_definitions/device-status.yaml
```diff
@@ -266,13 +773,54 @@ components:
...
-          schema:
-            $ref: "#/components/schemas/ErrorInfo"
-          example:
-            status: 504
-            code: "TIMEOUT"
-            message: "Request timeout exceeded. Try later."
+  examples:
+    ROAMING_STATUS:
+      value:
+        subscriptionId: 123654
```
fernandopradocabrillo, line 779: Hi! I'm new here so please ignore this if it causes any noise :D

Shouldn't the property in the examples also be `eventSubscriptionId`? It happens in all examples

code/API_definitions/device-status.yaml
JoachimDahlgren: Maybe we are saying the same thing but if I do not misunderstand the 3GPP standard you will get a notification only when changing from HPLMN to VPLMN and vice versa. You will not get a notification when moving from one VPLMN to another VPLMN. Even though there are implementations that will give you also notifications when changing between VPLMNs we cannot rely on it. Thereby the name of the event ROAMING_CHANGE_COUNTRY is a bit misleading as it indicates that you will always get such notifications.

eric-murray: Well, I think the use case for `ROAMING_CHANGE_COUNTRY` is clear - the API consumer will get a notification whenever the MCC of the PLMN ID of the network that the device is connected to changes.

As for getting this information, you are correct that the NEF will not generate a ROAMING_STATUS monitoringType event just because the device has changed VPLMN. However, that does not mean that the NEF cannot be used to get this information. The NEF can be regularly asked (i.e. polled) for a one-time report as to the current roaming status and PLMN ID. For this use case, the polling would not have to be very frequent, and would only have to be done for devices for which `ROAMING_CHANGE_COUNTRY` events have been requested.

And, as you mention above, there are proprietary implementations that can give this information. Indeed, we are using a proprietary implementation in Vodafone, and this use case is important to us. If any API provider does not want to implement `ROAMING_CHANGE_COUNTRY` events because the NEF does not directly give these notifications, then they are free to do that.

code/API_definitions/device-status.yaml
eric-murray: Thanks @fernandopradocabrillo 

@bigludo7 
Can you update the notification examples?

patrice-conil: Hi @eric-murray,
@bigludo7 is OoO, I made PR #64 to fix typos

BR,
Patrice

eric-murray: Thanks @patrice-conil. I merged these changes into the PR, but that invalidates previous approvals.

@sachinvodafone, @akoshunyadi, @JoachimDahlgren Can you review again?

@fernandopradocabrillo Can you review and approve for Telefonica?

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
sachinvodafone: Additionally, I have a basic question. In the case that a subscription has been deleted, will it still be included in the GET eventSubscription? If it is not included, it would restrict the client from accessing the subscription's history. Conversely, if it is included, we should specify whether a specific subscription has been deleted, even if its expiration time is still valid.

code/API_definitions/device-status.yaml
eric-murray: @fernandopradocabrillo
What you are missing is that the different notification types are something that the API consumer subscribes to (i.e. is an input when they create a subscription). So if an API consumer subscribed to `ROAMING_ON`, they would ONLY get a notification when the device moved from the HPLMN to a VPLMN.

To get the notifications in your example, you would need to subscribe to all three of the `ROAMING_ON`, `ROAMING_CHANGE_COUNTRY` and `ROAMING_OFF` events (i.e. have 3 separate subscriptions).

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
eric-murray: @sachinvodafone
I had anyway been assuming that `GET /event-subscriptions` would only return active subscriptions. What is the history that you think the API consumer might still need access to for an expired subscription, remembering that CAMARA APIs are designed to be called by code and not by people?

Providing details of expired subscriptions can anyway be implementation specific. But you would need a good argument if filtering options are to be included in the API design.

code/API_definitions/device-status.yaml
fernandopradocabrillo: Yes, that part understood! My question is, what is the difference between subscribing only to `ROAMING_STATUS` or subscribing to both `ROAMING_ON` and `ROAMING_OFF`? What are the benefits of subscribing to `ROAMING_STATUS` only apart of having to subscribe only to one event? 

In the current model, if I subscribe only to `ROAMING_ON` because I am only interested in that event, I do not receive information on which country is the SIM roaming, I am forced to subscribe also to `ROAMING_STATUS` to get that information.

And If we say that the `ROAMING_STATUS` represents a change in roaming status, why don't we include the country change in the actions that cover this event?

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
sachinvodafone: Hi @eric-murray Thanks for your notes. I was referring the notes between @bigludo7 and @monamok on the same topic so thought to clarify if we have reached to any alignment or not. 

Regarding whether subscription history (subscription ID , creation date, expire date, deletion status) is required or not, while not all API consumers may require access to historical data of expired subscriptions, there are scenarios where it can still be relevant and valuable for client.  For example: 

1. In **Audit and Compliance** where  it may be necessary to maintain a complete audit trail of all subscriptions, including expired/deleted ones.
2. In **Billing and Financial Records** where  access to these  data may be valuable for accounting, billing reconciliation, or financial reporting purposes.
3. In **Analysis and Reporting** where historical data on expired subscriptions may be useful for analytical purposes

code/API_definitions/device-status.yaml
eric-murray: The difference is that you only have one subscription to manage instead of two. The only reason to subscribe to both `ROAMING_ON` and `ROAMING_OFF` rather than `ROAMING_STATUS`, is because you need separate subscriptions, possibly with separate notification endpoints and certainly with different `eventSubscriptionId`. Otherwise, use `ROAMING_STATUS`.

There is an argument for adding roaming country for `ROAMING_ON` notifications, but I don't know the use case for those events. It may be that the use case only requires the API consumer to know that the device is now roaming, but it does not matter where. This is not a use case that Vodafone is interested in.

`ROAMING_STATUS` notifications can include the new country that triggered the event. Have a look at the example in the API definition. If a client wants to know what the previous country was, they need to store the previous notification.

code/API_definitions/device-status.yaml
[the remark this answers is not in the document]
eric-murray: The current API definition would allow `GET /event-subscriptions` to return expired subscriptions, but I don't see that any of the reasons you list would make this mandatory. API consumers can keep their own records of expired subscriptions (there is not that much data per subscription), and the API provider would anyway need to keep records for the reasons you list, but that does not mean they need to make them available to the API consumer through the CAMARA service API.

So I still don't see any reason to mandate that `GET /event-subscriptions` returns expired subscriptions, and hence no reason to add filtering options.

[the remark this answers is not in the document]
sachinvodafone: If we are sharing only active subscriptions, the description of the "ExpireAt" field should indicate the "date and time when the subscription will expire" instead of mentioning both "when the subscription will expire **or has expired**."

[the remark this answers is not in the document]
eric-murray: I already said that API providers can give details of expired subscriptions if they choose to implement that. But there is no reason to mandate it, nor to prohibit it.

API consumers should not rely on the API for historical records of their subscriptions. They need to be more organised than that.

code/API_definitions/device-status.yaml
sachinvodafone: Just checking , even if we include country information in the ROAMING_ON and ROAMING_OFF events, can't we control which clients are authorized to view the country code information and restrict it for those who don't have the necessary privileges . Sorry , if doesn't make sense.

fernandopradocabrillo: Okey then, fine from my side!

code/API_definitions/device-status.yaml
```diff
@@ -588,12 +777,50 @@ components:
...
-          roaming: true
-          countryCode: 208
-          countryName: 
-            FR 
+        event:
+          eventType: ROAMING_STATUS
+          eventTime: 2023-01-17T13:18:23.682Z
+          eventDetail:
+            device:
+              phoneNumber: 123456789
```
JoachimDahlgren, line 785: The example given for the events all include phoneNumber as example of the identification of the device. When we set up a subscription there are several ways to identify the device for which we want to subscribe to device status. All those being optional. 

I would assume that whatever you input as an identifier is also what you will receive back with the event. The reason for that would be that if the application have not been trusted with knowing the actual phoneNumber of the device this should not be exposed when calling the Device Status API. Is this a common understanding and in that case should it be mentioned in the documentation of the API?

eric-murray: Yes, that is my understanding - the API should return the at most same device identifiers as were provided by the client, but could maybe remove some if they were not valid or not used by the API provider.

For example, if the API client provides both `phoneNumber`  and `ipv4Address`, but of those two the implementation only uses `phoneNumber` to identify the device, then only `phoneNumber` should be included in responses. But that is my view - it would not be wrong to return `ipv4Address` as well even if that could be interpreted as confirming that the device is using that IP address when that has not been checked.

My view is that returning these values also gives the API provider an opportunity to "clean up" the formatting to show the API client the preferred field formatting. As we tighten up format patterns, there is less scope for the API client to use formats that are difficult to parse but, for example, if the client provided `+123456789` and the API returned `123456789`, that would indicate that we would prefer the optional `+` not to be used.

eric-murray (dismissed):

code/API_definitions/device-status.yaml
```diff
@@ -205,6 +466,244 @@ components:
...
+
+    SubscriptionEnds:
+      description: Event detail structure for SUBSCRIPTION_ENDS event
+      type : object
+      required:
+        - device
+      properties:
+        device:
+          $ref: '#/components/schemas/Device'
+
```
akoshunyadi, line 706: I think this eventDetail should contain the property terminationReason

eric-murray: This would be an additional property in the `SubscriptionEnds` schema.

But what reasons are there for the subscription ending other than it has expired (i.e. the `subscriptionExpireTime` has been reached)? I can see that there may be other reasons in theory, but I don't see that they will be encountered in practice. Why would the API provider choose to terminate the subscription early?

In any case, the API client can infer "other reasons" for subscription termination if `subscriptionExpireTime` has not been reached.

So I don't have a problem with adding a `terminationReason` property, but what would be the reasons other than `SUBSCRIPTION_EXPIRED`?

eric-murray: OK, fair point that this is a requirement of the CAMARA API Design Guidelines.

The example used in the design guidelines suggests that `terminationReason` is just a freeform string. So can we just add the property with no pre-defined values? Or should it be an enum, maybe {`SUBSCRIPTION_EXPIRED`, `NETWORK_TERMINATED`}?

akoshunyadi: Yes, it could be both... I tend to use enums, it's more uniform through all the APIs and can be evaluated programmatically

eric-murray: OK, I added the terminationReason as an enum

bigludo7:
As @fernandopradocabrillo steps in for @monamok I removed review request from Mona.

code/API_definitions/device-status.yaml
eric-murray: Maybe a topic for the TSC, but the point is that the current definition does allow the API consumer to provide more than one device identifier. This is inherited from QoD, where the motivation to allow more than one is that the device may be dual-stack, and we needed a way that the API consumer could specify if they wanted to implement the enhanced QoS for the IPv4 flow, IPv6 flow or both (providing just MSISDN or External Id would imply both if dual-stack).

Obviously this motivation does not apply here, but also I feel we should encourage the API consumer to provide the device identifiers they know, because some are easier to work with than others. It is much easier to determine the roaming status of a MSISDN than a public IP address, and if we restricted the number of device identifiers to one, they might just give us the IP address.

My feeling it is up to the API implementor how they process multiple device identifiers. They can do a consistency check if they want, or just take the identifier most convenient for their implementation. I don't really see that we need to standardise here. Device identifiers are inherently CSP specific.

eric-murray (approved):

bigludo7:
@fernandopradocabrillo May I merge or do you want to review?

fernandopradocabrillo (approved):
LGTM

Issue #14, opened 2023-01-12
Align error message with structure defined in the guideline


bigludo7:
Hello,
For consistency we should align the error structure with the one identified in the [Camara API design guideline ](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#6-error-responses) in order to provide an unified representation to developper community.
This should be aligned for future api release.

GarethWGSMA:
Yes I would agree. We need as much consistency, for the developers, as possible across all the api's.

NoelWirzius:
We should follow the guidelines when they are aligned

monamok:
I opened [PR#35](https://github.com/camaraproject/DeviceStatus/pull/35) to adjust the schema references in the errors responses and apply the agreed solution: [Issue](https://github.com/camaraproject/WorkingGroups/issues/151).

It also edites some error codes to be aligned with those described in [this section of the guidelines](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-design-guidelines.md#6-error-responses).

I think it makes sense to have a 504/Timeout error in this API so I added it in. If there is any comments please feel free to share it.

bigludo7:
Implemented !

PR #35, opened 2023-03-29
Errors definition alignment with guidelines

monamok:
This PR is opened to handle [issue#14](https://github.com/camaraproject/DeviceStatus/issues/14). 

Besides fixing error codes & error model, additionally I add a new error 504/Timeout that I think it could be useful. 
Looking forward to getting your feedback.


bigludo7 (approved):
Thanks you Mona !!
Look very good for me.

eric-murray (approved):

PR #36, opened 2023-03-31
MoM 2023-03-29

akoshunyadi:


bigludo7 (approved):
Thanks !
Look good for me.
@akoshunyadi sorry to jump to another topic but  I would like to set up a call to discuss Notification with you and Pedro - May I get your email (mine: ludovic.robert@orange.com) . Thanks

eric-murray (approved):

PR #44, opened 2023-05-03
Update Connectivity_API.md

akoshunyadi:
#### What type of PR is this?

* documentation

This PR updates the API documentation to 0.4.1.

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/API_documentation/Connectivity_API.md
+++ b/documentation/API_documentation/Connectivity_API.md
@@ -5,10 +5,9 @@ This API makes possible to query the connectivity status of user equipment.
 <img src="resources/DeviceStatus_Connectivity_diagram.png">
 
 ## 2\. Quick Start
-DeviceStatus API v0.2.0 exposes only one endpoint, which can be used to check connectivity status for user equipment. To do so user has to pass following parameters in request body:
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Thanks @akoshunyadi 
Seems to me that we have few discrepancies between the error listed here and the one in the swagger/guidelines ,
401 should be UNAUTHENTICATED and not Unauthorized
etc...

eric-murray:
@bigludo7
The documentation is specific to [v0.4.1](https://github.com/camaraproject/DeviceStatus/blob/release-0.4.1/code/API_definitions/device-status.yaml), and that version didn't follow the design guidelines very well. Some clean up of the error messages was done for 0.5.0-wip, so the next version should be better.

Although this documentation doesn't quite match the error examples given in v0.4.1, it's close enough and I don't see any great value in spending additional effort documenting more exactly something that will anyway be changing soon when v0.5.0 is released.

So I'm tempted to just merge this, consider v0.4.1 finished and then move on with v0.5.0. Let me know what you think.

bigludo7:
@eric-murray Got it !
Thanks Eric.
Works for me.

eric-murray (approved):

PR #65, opened 2023-07-19
Add documentation

### LINKED ISSUE #57: Add documentation into API definition
**Problem description**
The documentation in the API definition should be extended so, that no extra documentation is needed.

**Expected action**
Add documentation to API definition as discussed in https://github.com/camaraproject/WorkingGroups/issues/164

**Additional context**

bigludo7:
Add documentation for the API
Solve issue #57

#### What type of PR is this?

* documentation

This PR add the documentation in the API swagger.

Fixes #57 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -46,7 +46,7 @@ info:
 
     ## Device status event subscription
 
-    This endpoint allows the API consumer to create and delete event subscriptions for specific roaming device status events. The CAMARA subscription model is detailed in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md).
+    This endpoint allows the API consumer to create and delete event subscriptions for specific roaming device status events. The CAMARA subscription model is detailed in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#121-subscription).
 
     Following `eventType` options are defined for this API: 
       - `ROAMING_STATUS` - This event is triggered when the device roaming status changes from roaming ON (`true`) to roaming OFF (`false`) or vice-versa
@@ -59,7 +59,7 @@ info:
 
     ## Notification callback
 
-    This endpoint describes the event notification that will be received by the subscription listener (API consumer) side when a subscribed event has occurred. A detailed description of the event notification mechanism is provided in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md).
+    This endpoint describes the event notification that will be received by the subscription listener (API consumer) side when a subscribed event has occurred. A detailed description of the event notification mechanism is provided in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#122-event-notification).
 
     **WARNING**: This endpoint must be implemented on the listener (API consumer) side, as a `POST /notifications` request will be made by the device status resource server to the url provided by notification listener.
 
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -2,60 +2,66 @@ openapi: 3.0.3
 info:
   title: Device Status
   description: |
-    This API provides the customer with the ability to query device roaming status and subscribe to event related to device roaming status.
+    This API provides the customer with the ability to query the roaming status of a device, and to subscribe to receive notifications when the device roaming status changes.
 
     # Introduction
 
-    API consumer is able to verify whether a certain user device is in roaming situation (or not). This capability is provided in 2 ways:
-      - via direct request with the roaming situation in the response.
-      - via a subscription  request - in this case the roaming situation is not in the response but event notification is sent back to the event subscriber when roaming situation change.
+    The API consumer is able to verify whether a certain user device is in a roaming situation (or not). This capability is provided in two ways:
+      - via a direct request, with the current roaming status given in the response.
+      - via a subscription request - in this case the roaming situation is not in the response but a notification is sent back to the subscriber when the device roaming status changes
 
-    The verification of the roaming situation depends on the network's ability. Additionally to the roaming status, when the device is in roaming situation, visited country information is send back in the response.
+    The verification of the roaming situation depends on the network's ability. In addition to the roaming status, visited country information is included in the response when the device is roaming.
 
-    Device status verification could be useful in scenario such as (not exhaustive):
-      - For regulatory reasons, where a customer may need to be within a certain jurisdiction, or out with others, in order for transactions to be authorized
-      - For security / fraud reasons, to establish that a customer is located where they claim to be
+    Device roaming status verification could be useful in scenarios such as (not exhaustive):
+      - For regulatory reasons, where a customer may need to be within a certain jurisdiction, or outwith others, in order for transactions to be authorized
+      - For security / fraud / fair play reasons, to establish that a customer is located where they claim to be
       - For service delivery reasons, to ensure that the customer has access to particular service, and will not incur roaming charges in accessing them
 
 
     # Relevant terms and definitions
 
-    * **Device**: A device refers to any physical entity that can connect to a network and participate in network communication.
-      At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
+    * **Device**: A device (i.e. user equipment) refers to any physical entity that can connect to a mobile network and participate in network communication.
+      At least one identifier for the device must be provided from the following four options:
+      - IPv4 address
+      - IPv6 address
+      - Phone number (MSISDN)
+      - Network Access Identifier assigned by the mobile network operator for the device
 
-    * **Roaming** : Roaming status. `true`, if device is in roaming situation - `false` else.
+    * **Roaming** : Roaming status:
+      - `true` if device is connected to a network other than its home network
+      - `false` if connected to its home network
 
-    * **Country** : Country code and name - visited country information, provided if the device is in roaming situation.
+    * **Country** : Visited country code and name:
+      - Mobile country code (MCC), as defined by ITU-T E.212 (e.g. `262`).
+      - ISO country code, as defined by ISO 3166 ALPHA-2 (e.g. `DE`). A list may be provided if multiple countries share the same MCC.
 
 
     # API Functionality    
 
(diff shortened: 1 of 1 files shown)


eric-murray (dismissed):

Sachinsiso (approved):

fernandopradocabrillo (commented):
I think we should include documentation regarding the use of the sync/async use of the API. Something along the lines of indicating that the response code will be implementation-dependent.
And I would complete the async behaviour a bit more by pointing out that a GET request to /event-subscriptions/{eventSubscriptionId} should be made after the creation/deletion of the subscription to check if the operation was successful.

eric-murray:
@fernandopradocabrillo
Can you propose some additional text? I don't think we need to document optional behaviour on the API consumer's part.

bigludo7:
@fernandopradocabrillo (additionnally to @eric-murray comment)
Fair point but one question for you before to update this doc: this point shouldn't be documented in the design guideline API ? As it should be common for all subscriptions it seems to me that is should be there. Thoughts?

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,72 @@
...
+
+
+    # Relevant terms and definitions
+
+    * **Device**: A device (i.e. user equipment) refers to any physical entity that can connect to a mobile network and participate in network communication.
+      At least one identifier for the device must be provided from the following four options:
+      - IPv4 address
+      - IPv6 address
+      - Phone number (MSISDN)
+      - Network Access Identifier assigned by the mobile network operator for the device
```
JoachimDahlgren, line 28: Should we maybe add a note that when you receive **Device** in a response or notification you will not get more device identifiers back than you used in the request?

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,72 @@
...
+    This endpoint allows the API consumer to create and delete event subscriptions for specific roaming device status events. The CAMARA subscription model is detailed in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md).
+
+    Following `eventType` options are defined for this API: 
+      - `ROAMING_STATUS` - This event is triggered when the device roaming status changes from roaming ON (`true`) to roaming OFF (`false`) or vice-versa
+
+      - `ROAMING_ON` - This event is only triggered when the device changes from roaming OFF to roaming ON
+
+      - `ROAMING_OFF` - This event is only triggered when the device changes from roaming ON to roaming OFF
+
+      - `ROAMING_CHANGE_COUNTRY` - This event is triggered whenever the mobile country code (MCC) of the network to which the device is connected changes. Changing netorks within a given country does not trigger this event.
```
JoachimDahlgren, line 58: Small typo missing a "w" in networks.

JoachimDahlgren (changes requested):

fernandopradocabrillo:
> @fernandopradocabrillo (additionnally to @eric-murray comment) Fair point but one question for you before to update this doc: this point shouldn't be documented in the design guideline API ? As it should be common for all subscriptions it seems to me that is should be there. Thoughts?

@bigludo7 Yes! I think it definitely should be well documented in the guidelines since it is going to be the same behaviour for all the APIs. But, don't you think we should mention it briefly here too? To make it easier for the implementer to understand how it works.

> @fernandopradocabrillo Can you propose some additional text? I don't think we need to document optional behaviour on the API consumer's part.

@eric-murray I was considering that it would be better if the documentation reflects the full behaviour of the API and what should be expected in each scenario. In this case indicating that a GET must be done to ensure the existence or successful deletion of the subscription is to point out that for asynchronous scenarios, there is no other way to check that the operation has been performed successfully. What do you think?

code/API_definitions/device-status.yaml
eric-murray: Fixed

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,72 @@
...
+
+    The API exposes following capabilities:
+
+    ## Device roaming status
+
+    The endpoint `POST /roaming` allows the APi consumer to synchronously (i.e. immediately) get roaming status and country information (if the device is roaming).
+
+    ## Device status event subscription
+
+    This endpoint allows the API consumer to create and delete event subscriptions for specific roaming device status events. The CAMARA subscription model is detailed in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md).
```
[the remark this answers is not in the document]
eric-murray: I updated the links to be hash-links to the most appropriate numbered section in the design guidelines. Unfortunately the sections on instance-based and resource-based subscriptions are not numbered, so cannot be directly hash-linked. This could be changed.

My preference is not to copy over information from the design guidelines, because then we have a maintenance issue whenever the design guidelines are updated. Of course, by hash-linking there is still an issue if the section names or numbers change, but probably more manageable.

Anyway, that's my proposal, but happy to discuss.

code/API_definitions/device-status.yaml
eric-murray: I added a note to this effect

eric-murray:
@fernandopradocabrillo
I agree with @bigludo7 that details around the operation of the subscription / notification mechanism are best put in the API design guidelines themselves and linked to from the documentation here. So I'd suggest to modify that document, as the behaviour applies to all resource-based subscriptions, and the Device Status API will pick up on the documentation through the hash-links.

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,72 @@
...
+      - `ROAMING_OFF` - This event is only triggered when the device changes from roaming ON to roaming OFF
+
+      - `ROAMING_CHANGE_COUNTRY` - This event is triggered whenever the mobile country code (MCC) of the network to which the device is connected changes. Changing networks within a given country does not trigger this event.
+
+    ## Notification callback
+
+    This endpoint describes the event notification that will be received by the subscription listener (API consumer) side when a subscribed event has occurred. A detailed description of the event notification mechanism is provided in the [CAMARA API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#122-event-notification).
+
+    **WARNING**: This endpoint must be implemented on the listener (API consumer) side, as a `POST /notifications` request will be made by the device status resource server to the url provided by notification listener.
+
```
akoshunyadi, line 65: Is /notifications mandatory?
Doesn't the webhook/notificationUrl define the complete url?

eric-murray: Hi @akoshunyadi 

Good question. I had assumed that it was mandatory for the notification endpoint name to match the name defined in the API definition. So either the API consumer includes that in `notificationUrl`, or it gets added on automatically to whatever is specified.

However, looking at the example in the [CAMARA Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#12-subscription-notification--event), I see that the API consumer specified `"notificationUrl": "https://application-server.com"`but the notification was sent to `https://application-server.com/v0/notifications`.

So now I'm asking "where did the `v0` come from?". Rules around constructing the full notification URL need to be clarified.

But, despite that, my assumption was that if the API consumer had instead specified `"notificationUrl": "https://application-server.com/notification"`, then that is exactly where the notification would be sent (rather than `"https://application-server.com/notification/notification"`).

Sachinsiso: As per my understanding, when the API consumer registers its webhook/notification URL with the API provider, the provider should perform validation on the URL to ensure it meets the required format that should be properly documented at our end. If there are any issues, such as missing components or incorrect structure (missing “/notification”), the API provider should respond with an appropriate error message indicating what went wrong and how the API consumer can correct the URL.

Also it may create some confusion to have extra endpoint in our swagger file as  if client try to access “Try it out” feature then this extra endpoint will take him/her to  http://localhost:9091//device-status/v0/notifications) that can confuse him/her.

akoshunyadi: In QoD we decided not to add or expect "/notifications", just use the provided callback url
notifications:
          "{$request.body#/webhook/notificationUrl}":

eric-murray: Given that we have an explicit `/notifications` endpoint defined here, and given the resource-based subscription example used in the design guidelines, I think we have to assume that the notification endpoint must end with `/notifications`.

The question is what we should do if the `notificationUrl` provided by the API consumer does not end with `/notifications`? The example in the design guidelines suggests (more or less) that we should just add `/notifications` to the endpoint provided by the API consumer.

But I agree with @Sachinsiso that a better option might be to reject it and get the API consumer to try again to avoid the risk that the API consumer has misunderstood how to implement this endpoint. This interpretation would require a change in the design guidelines.

I'd suggest to wait until @bigludo7 returns before discussing further, and keep this PR open in the meantime.

PedroDiez: Hi,

will also align this with @fernandopradocabrillo, anycase I think it is good to wait until Ludovic returns. 

I consider the comment is valid, and probably this can generate misunderstanding. When sending notification from Telco Operator to client, the endpoint is settled by {$request.body#/webhook/notificationUrl}, so really not required any append like "/notifications". The notification is sent to the endpoint provided by the client.

Think it is good time to have this discussion given there is an issue where a "traversal" API is being discussed/defined to cover this flow: https://github.com/camaraproject/Commonalities/issues/8

bigludo7:
I've identified here 2 topics all related to events:

**1.** use of `callbacks`: OAS definition (like in QoD API) or add a specific `/notification` endpoint  (like we have there). I guess the decision should be at CAMARA level and not per API. @PedroDiez  raised issue https://github.com/camaraproject/Commonalities/issues/44 to track this.
**2.** definition of notification url - I tend to think that we need to fix this and proposal by @PedroDiez to use {$request.body#/webhook/notificationUrl} as it is defined in QoD v0.0.9 seems to be a good approach.

Additionally there is a third point (**3.**) raised in commonalities that could impact this API: should we keep current notification model (https://github./camaraproject/Commonalities/pull/41) or move to cloud event notification (https://github.com/camaraproject/Commonalities/pull/43).

I suggest to wait until commonalities outputs for 1. & 3. before to move forward for this PR.

bigludo7 (commented):
To provide clarification about Notification update
Aligned with https://github.com/camaraproject/Commonalities/pull/56

bigludo7:
I propose to close this issue as the discussion is not about the documentation but to the shift to CloudEvent. I will open another PR with the new yaml proposal.

bigludo7:
New PR will be created for #74

PR #69, opened 2023-08-15
add /connectivity - endpoint to check the network status of a device

### LINKED ISSUE #34: New Endtype Connectivity
**Summary**
===========
In this issue will add a new **Endpoint** which helps to get the current connectivity status. 

**Changes** 
==========
* Adding new Endpoint
    * /connectivity
      * This will give back if the device is currently connected to a network or not (DATA)

maxl2287:
#### What type of PR is this?

* enhancement/feature

This PR creates a new endpoint "/connectivity" to check the current network status of a device.

Fixes #34

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -5,6 +5,8 @@ info:
     This API provides the customer with the ability to query device: 
       - Roaming Status
       - Connectivity Status
+      
+    
     Moreover, this API extends the functionality by allowing users customers to subscribe to events associated with these status queries.
 
     # Introduction
@@ -39,9 +41,9 @@ info:
     * **Country** : Country code and name - visited country information, provided if the device is in roaming situation.
     
     * **Connectivity** : Connectivity status.
-      - 'CONNECTED_SMS', if device is connected to the network via SMS usage 
-      - 'CONNECTED_DATA', if device is connected to the network via data usage' 
-      - 'NOT_CONNECTED', if device is not connected to the network'
+      - `CONNECTED_SMS`, if device is connected to the network via SMS usage 
+      - `CONNECTED_DATA`, if device is connected to the network via data usage' 
+      - `NOT_CONNECTED`, if device is not connected to the network'
 
     # API Functionality    
 
@@ -478,6 +480,7 @@ components:
         CONNECTED_SMS: The device is connected to the network for SMS usage
 
         NOT_CONNECTED: The device is not connected
+      type : string
       enum:
         - CONNECTED_DATA
         - CONNECTED_SMS
@@ -653,7 +656,7 @@ components:
         device:
           $ref: "#/components/schemas/Device"
         type:
-          $ref: "#/components/schemas/SubscriptionEventType"
+          $ref: "#/components/schemas/SubscriptionCreationEventType"
 
     SubscriptionEventType:
       type: string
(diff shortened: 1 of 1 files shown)


maxl2287:
Hi @eric-murray,

yes I thought so, too.
That's why I updated the version to 0.6.0-wip.

bigludo7:
@maxl2287  Agree with @eric-murray that we must first stabilize the v0.5 with the notification and the proceed this PR in v0.6. 
Hope we will be able to solve quickly the notification.

One question - Why did you removed CONNECTED_DATA_SMS value ? (the device is connected to the network for Data & SMS usage). Sorry if i missed a discussion on this topic but IMHO it makes sense to keep the value when the device can receive Data &SMS. Thanks

maxl2287:
> One question - Why did you removed CONNECTED_DATA_SMS value ? (the device is connected to the network for Data & SMS usage). Sorry if i missed a discussion on this topic but IMHO it makes sense to keep the value when the device can receive Data &SMS. Thanks

I thought that we can remove it, because it's already inside of CONNECTED_DATA. 
Why? - Because I do not know a usecase, where you are connected to data only without be able to receive SMS.
So CONNECTED_DATA_SMS is for me the same like CONNECTED_DATA.

What do you think about this?

bigludo7:
Thanks @maxl2287 for the quick answer. Got it. I will check internally with my network experts but makes sense to me now.

akoshunyadi:
Decision from the DeviceStatus working group call: this PR should be extended by the connectivity events. However they should already be based on CloudEvents. To avoid conflicts this PR should be based on PR #74 .

maxl2287:
@eric-murray & @bigludo7: 

Depending on the MeetingMinutes (https://github.com/camaraproject/DeviceStatus/blob/main/documentation/MeetingMinutes/MoM-2023-09-13.md)
this PR needs to be part of 0.5.0 right?

I would like to wait until PR #75 is done to do the updates for this PR.

code/API_definitions/device-status.yaml
```diff
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.5.0-wip
+  version: 0.6.0-wip
```
sachinvodafone, line 11: Under description , we can mention "connectivity" too like "This API allows to query device roaming status, **connectivity** and subscribe to event related to device roaming status."

maxl2287:
@sachinvodafone @eric-murray @bigludo7 
I have updated this PR based on CloudEvents.

Please review. :)

bigludo7 (commented):
Hello @maxl2287 
Thanks for the proposal.
Few cosmetic change and one code change proposal: The eventType org.camaraproject.device-status.v0.subscription-ends has not to be subscribed (see API description). This notification does not require dedicated subscription. But, the eventType could be send in the notification so I've introduced 2 definition of eventType: one with and one without subscription-ends. 

Hope it makes sense for you

bigludo7:
@maxl2287 This is the same comment for https://github.com/camaraproject/DeviceLocation/issues/113 ;)

bigludo7:
> ```yaml
> SubscriptionDetail
> ```

Hello @maxl2287 
Yes. Sorry for the proposal - we can simplify.

My only request is to have a distinct array for the `type` attribute in `SubscriptionDetail` from the array in `CloudEvent`.

maxl2287:
@bigludo7 sorry I have inserted the wrong code as a proposal 😄 .

I mean: 
Is this here a good way to reduce redundancy:

```yaml
    SubscriptionEventType:
      allOf:
        - $ref: "#/components/schemas/SubscriptionCreationEventType"
        - type: string
          description: |
            subscription-ends: Event triggered when the subscription ends.
          enum:
            - org.camaraproject.device-status.v0.subscription-ends
```

bigludo7:
> @bigludo7 sorry I have inserted the wrong code as a proposal 😄 .
> 
> I mean: Is this here a good way to reduce redundancy:
> 
> ```yaml
>     SubscriptionEventType:
>       allOf:
>         - $ref: "#/components/schemas/SubscriptionCreationEventType"
>         - type: string
>           description: |
>             subscription-ends: Event triggered when the subscription ends.
>           enum:
>             - org.camaraproject.device-status.v0.subscription-ends
> ```

Yes this is excellent. I should confess that I did not know that it was possible to extend an enum this way. Thanks !!

bigludo7 (dismissed):
LGTM
Thanks @maxl2287

sachinvodafone (dismissed):

maxl2287:
> > @bigludo7 sorry I have inserted the wrong code as a proposal 😄 .
> 
> > 
> 
> > I mean: Is this here a good way to reduce redundancy:
> 
> > 
> 
> > ```yaml
> 
> >     SubscriptionEventType:
> 
> >       allOf:
> 
> >         - $ref: "#/components/schemas/SubscriptionCreationEventType"
> 
> >         - type: string
> 
> >           description: |
> 
> >             subscription-ends: Event triggered when the subscription ends.
> 
> >           enum:
> 
> >             - org.camaraproject.device-status.v0.subscription-ends
> 
> > ```
> 
> 
> 
> Yes this is excellent. I should confess that I did not know that it was possible to extend an enum this way. Thanks !!

I don't know. I thought so, but in DeviceStatus @jlurien denied that possibility.

So I guess I need to revert it.

maxl2287:
@bigludo7 I have reverted my commit and deleted an unused component.
Now I think its good to be reviewed.

I was not aware that an "extend" is not possible for enums yet and it's only possible by a bit of redundancy.

bigludo7 (approved):
LGTM. Thanks @maxl2287

sachinvodafone (approved):

PR #75, opened 2023-09-27
Added Notification management aligned with CloudEvents format

### LINKED ISSUE #74: Align event-subscriptions with updated design guidelines based on CloudEvents
**Problem description**
After approving [PR56 Update DG with use of callbacks & cloudEvents](https://github.com/camaraproject/Commonalities/pull/56) the DeviceStatus API definition needs to be updated too.

**Expected behavior**
Update the event-subscription related parts in the API spec based on the updated DG

**Alternative solution**

**Additional context**

bigludo7:
#### What type of PR is this?

* enhancement/feature

Aligned with CloudEvents format as defined in Commonalities here: https://github.com/camaraproject/Commonalities/pull/56 

Fixes #74 #57 

#### Changelog input

```
- Provide roaming status subscription & notification aligned with Cloudevents structure.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -898,7 +898,7 @@ components:
     ROAMING_CHANGE_COUNTRY:
       value:
         id: 123657
-        source: supertelco.notificationSendServer12
+        source: https://notificationSendServer12.supertelco.com
         type: org.camaraproject.device-status.v0.roaming-change-country
         specversion: "1.0"
         datacontenttype: application/json
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -912,7 +912,7 @@ components:
     SUBSCRIPTION_ENDS:
       value:
         id: 123658
-        source: supertelco.notificationSendServer12
+        source: https://notificationSendServer12.supertelco.com
         type: org.camaraproject.device-status.v0.subscription-ends
         specversion: "1.0"
         datacontenttype: application/json
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -1,7 +1,73 @@
 openapi: 3.0.3
 info:
   title: Device Status
-  description: Service Enabling Network Function API for monitoring device status. This API allows to query device roaming status and subscribe to event related to device roaming status.
+  description: |
+    This API provides the customer with the ability to query device roaming status and subscribe to event related to device roaming status.
+
+    # Introduction
+
+    API consumer is able to verify whether a certain user device is in roaming situation (or not). This capability is provided in 2 ways:
+      - via direct request with the roaming situation in the response.
+      - via a subscription  request - in this case the roaming situation is not in the response but event notification is sent back to the event subscriber when roaming situation change.
+
+    The verification of the roaming situation depends on the network's ability. Additionally to the roaming status, when the device is in roaming situation, visited country information is send back in the response.
+
+    Device status verification could be useful in scenario such as (not exhaustive):
+      - For regulatory reasons, where a customer may need to be within a certain jurisdiction, or out with others, in order for transactions to be authorized
+      - For security / fraud reasons, to establish that a customer is located where they claim to be
+      - For service delivery reasons, to ensure that the customer has access to particular service, and will not incur roaming charges in accessing them
+
+
+    # Relevant terms and definitions
+
+    * **Device**: A device refers to any physical entity that can connect to a network and participate in network communication.
+      At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
+
+    * **Roaming** : Roaming status. `true`, if device is in roaming situation - `false` else.
+
+    * **Country** : Country code and name - visited country information, provided if the device is in roaming situation.
+
+
+    # API Functionality    
+
+    The API exposes following capabilities:
+
+    ## Device roaming situation
+
+    The endpoint `POST /roaming` allows to get roaming status and country information (if device in roaming situation) synchronously.
+
+    ## Device status subscription
+
+    Theses endpoints allow to manage event subscription on roaming device status event. 
+    The CAMARA subscription model is detailed in the CAMARA API design guideline document and follows Cloudevents specification.
+
+    It is mandatory in the subscription to provide the event `type` subscribed are several are managed in this API.
+
+    Following event ``type`` are managed for this API: 
+      - roaming-status - Event triggered when the device switch from roaming ON to roaming OFF and conversely
+
+      - roaming-on - Event triggered when the device switch from roaming OFF to roaming ON
+
+      - roaming-off: Event triggered when the device switch from roaming ON to roaming OFF
+
+      - roaming-change-country: Event triggered when the device in roaming change country code
+
+    Note: Additionnally to these `type`, notification are send also when the subscription ends. 
+    This notification did not requires dedicated subscription. 
+    It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely. 
+
+    ### Notification callback
+
+    This endpoint describes the event notification received on subscription listener side when the event occurred. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
+
+    **WARNING**: This callback endpoint must be exposed on the listener side as `POST /{$request.body#/webhook/notificationUrl}`
+      will be posted by the device status resource server to the webhook url provided by notification 
+      listener.
+
+    ## Further info and support
+
+    (FAQs will be added in a later version of the documentation)
+
   termsOfService: http://swagger.io/terms/
   contact:
     email: project-email@sample.com
@@ -17,9 +83,9 @@ security:
   - oAuth2ClientCredentials: []
   #  - BasicAuth: []
   #  - apiKey: []
-  - three_legged:
-      - device-status-roaming-read
-      - device-status-roaming-event
+  - threeLegged:
+      - device-status:roaming:read
+      - device-status:roaming-event-subscriptions:create
 servers:
   - url: '{apiRoot}/{basePath}'
     variables:
@@ -32,57 +98,10 @@ servers:
 tags:
   - name: Device roaming status
     description: Operation to get device roaming status and country information (if roaming) synchronously
-  - name: Device status event subscription
+  - name: Device status subscription
     description: Operation to manage event subscription on device status event (roaming, connectivity)
-  - name: Notification Callback
-    description: |
-      Notification received on subscription listener side. 
-      
-      **WARNING**: This endpoint must be exposed on the listener side as POST /notifications 
-      will be posted by the device status resource server to the url provided by notification 
-      listener.
+
 paths:
-  /notifications:
-    post:
-      tags:
-        - Notification Callback
-      summary: "Notification endpoint to notify listener that a subscribed event occurred"
-      description: Notification endpoint to notify listener that a subscribed event occurred. This endpoint must be exposed on the listener side.
-      operationId: postNotification
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/Notification"
-            examples:
-              ROAMING_STATUS:
-                $ref: '#/components/examples/ROAMING_STATUS'
-              ROAMING_CHANGE_COUNTRY:
-                $ref: '#/components/examples/ROAMING_CHANGE_COUNTRY'
-              ROAMING_ON:
-                $ref: '#/components/examples/ROAMING_ON'
-              ROAMING_OFF:
-                $ref: '#/components/examples/ROAMING_OFF'
-              SUBSCRIPTION_ENDS:
-                $ref: '#/components/examples/SUBSCRIPTION_ENDS'
-      responses:
-        204:
-          description: No Content
-        400:
-          $ref: "#/components/responses/Generic400"
-        401:
-          $ref: "#/components/responses/Generic401"
-        403:
-          $ref: "#/components/responses/Generic403"
-        500:
-          description: Server error
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/ErrorInfo"
-        503:
-          $ref: "#/components/responses/Generic503"
   /roaming:
     post:
       tags:
@@ -90,6 +109,9 @@ paths:
       summary: 'Get the current roaming status and the country information'
       description: Get the current roaming status and the country information
       operationId: getRoamingStatus
+      security:
+        - threeLegged: 
+          - device-status:roaming:read
       requestBody:
         content:
           application/json:
@@ -131,32 +153,81 @@ paths:
           $ref: "#/components/responses/Generic500"
         503:
           $ref: "#/components/responses/Generic503"
-  /event-subscriptions:
+  /subscriptions:
     post:
       tags:
-        - Device status event subscription
+        - Device status subscription
       summary: 'Create a device status event subscription for a device'
       description: Create a device status event subscription for a device
-      operationId: createDevicStatusEventSubscription
+      operationId: createDevicStatusSubscription
+      security:
+        - threeLegged:
+          - device-status:roaming-subscriptions:create
       requestBody:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/CreateEventSubscription'
+              $ref: '#/components/schemas/CreateSubscription'
         required: true
+      callbacks:
+        notifications:
+          "{$request.body#/webhook/notificationUrl}":
+            post:
+              tags:
+                - Session notifications callback
+              summary: "Session notifications callback"
+              description: |
+                Important: this endpoint is to be implemented by the API consumer.
+                The Device status server will call this endpoint whenever any device status related event occurs.
+              operationId: postNotification
+              requestBody:
+                required: true
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/CloudEvent"
+                    examples:
+                      ROAMING_STATUS:
+                        $ref: '#/components/examples/ROAMING_STATUS'
+                      ROAMING_CHANGE_COUNTRY:
+                        $ref: '#/components/examples/ROAMING_CHANGE_COUNTRY'
+                      ROAMING_ON:
+                        $ref: '#/components/examples/ROAMING_ON'
+                      ROAMING_OFF:
+                        $ref: '#/components/examples/ROAMING_OFF'
+                      SUBSCRIPTION_ENDS:
+                        $ref: '#/components/examples/SUBSCRIPTION_ENDS'
+              responses:
+                "204":
+                  description: Successful notification
+                "400":
+                  $ref: "#/components/responses/Generic400"
+                "401":
+                  $ref: "#/components/responses/Generic401"
+                "403":
+                  $ref: "#/components/responses/Generic403"
+                "500":
+                  $ref: "#/components/responses/Generic500"
+                "503":
+                  $ref: "#/components/responses/Generic503"
+              security:
+                - { }
+                - notificationsBearerAuth: [ ]
+
+
       responses:
         201:
           description: Created
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/EventSubscriptionInfo'
+                $ref: '#/components/schemas/SubscriptionInfo'
         202 :
           description: Request accepted to be processed. It applies for async creation process.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/EventSubscriptionAsync'
+                $ref: '#/components/schemas/SubscriptionAsync'
         400:
           $ref: '#/components/responses/Generic400'
         401:
@@ -171,10 +242,10 @@ paths:
           $ref: '#/components/responses/Generic503'
     get:
       tags:
-        - Device status event subscription
+        - Device status subscription
       summary: 'Retrieve a list of device status event subscription'
       description: Retrieve a list of device status event subscription(s)
-      operationId: retrieveEventSubscriptionList
+      operationId: retrieveSubscriptionList
       responses:
         200:
           description: List of event subscription details
@@ -183,7 +254,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/EventSubscriptionInfo'
+                  $ref: '#/components/schemas/SubscriptionInfo'
         400:
           $ref: '#/components/responses/Generic400'
         401:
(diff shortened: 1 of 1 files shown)


akoshunyadi:
@bigludo7 This PR fixes also Issue #57 ?

bigludo7:
> @bigludo7 This PR fixes also Issue #57 ?

@akoshunyadi  correct !

bigludo7 (commented):
Aligned with merged DG for event & subscriptions
@akoshunyadi @sachinvodafone

code/API_definitions/device-status.yaml
```diff
@@ -131,32 +153,81 @@ paths:
...
+                - Session notifications callback
+              summary: "Session notifications callback"
+              description: |
+                Important: this endpoint is to be implemented by the API consumer.
+                The Device status server will call this endpoint whenever any device status related event occurs.
+              operationId: postNotification
+              requestBody:
+                required: true
+                content:
+                  application/json:
```
jlurien, line 186: We may need to use `application/cloudevents+json` as stated in
https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md#3-envelope

bigludo7: @jlurien Yes make sense.
I will apply.

Do you think we need to go on this detail in the design guideline document as well? I tend to think yes.

bigludo7 (commented):
Changed` content: application/json`  to `content: application/cloudevents+json` for alignement with CloudEvents specification

code/API_definitions/device-status.yaml
jlurien: yes, I think it will be useful to add this and maybe provide a template or artifact as example. I learned about this from some @patrice-conil 's PR indeed

code/API_definitions/device-status.yaml
```diff
@@ -131,32 +153,81 @@ paths:
...
-  /event-subscriptions:
+  /subscriptions:
     post:
       tags:
-        - Device status event subscription
+        - Device status subscription
       summary: 'Create a device status event subscription for a device'
       description: Create a device status event subscription for a device
-      operationId: createDevicStatusEventSubscription
+      operationId: createDevicStatusSubscription
```
maxl2287, line 162: typo:
createDevic**e**StatusSubscription

code/API_definitions/device-status.yaml
```diff
@@ -546,114 +638,119 @@ components:
...
+    CloudEvent:
+      description: The notification callback
       required:
-        - eventType
-        - eventTime
-        - eventDetail
+        - id
+        - source
+        - specversion
+        - type
```
maxl2287, line 659: "time" is missing as required.

See Commonalities: 

```
Note: Attribute time is tagged as optional in CloudEvents specification but from CAMARA perspective we mandate to value this attributes.

https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#event-notification-definition

```

maxl2287 (changes requested):

code/API_definitions/device-status.yaml
```diff
@@ -779,53 +876,66 @@ components:
...
+        type: org.camaraproject.device-status.v0.roaming-status
+        specversion: "1.0"
+        datacontenttype: application/json
+        data: 
+          device:
               phoneNumber: 123456789
-            roaming: true
-            countryCode: 208
-            countryName: FR
+          roaming: true
```
sachinvodafone, line 888: As per the document, it's mentioned that "The event subscription identifier must have a value when dealing with Resource-based subscriptions." So we should include the "**subscriptionId**" in the callback response if it's relevant or applicable in that context.

bigludo7 (commented):
Took @maxl2287 and @sachinvodafone review comments into account;
Thanks for the review.

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,73 @@
...
+      - ``org.camaraproject.device-status.v0.roaming-status`` - Event triggered when the device switch from roaming ON to roaming OFF and conversely
+
+      - ``org.camaraproject.device-status.v0.roaming-on`` - Event triggered when the device switch from roaming OFF to roaming ON
+
+      - ``org.camaraproject.device-status.v0.roaming-off``: Event triggered when the device switch from roaming ON to roaming OFF
+
+      - ``org.camaraproject.device-status.v0.roaming-change-country``: Event triggered when the device in roaming change country code
+
+    Note: Additionnally to these list, ``org.camaraproject.device-status.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
+    This notification did not requires dedicated subscription.
```
akoshunyadi, line 56: Typo: did not requires => does not require

code/API_definitions/device-status.yaml
```diff
@@ -131,50 +153,98 @@ paths:
...
+                The Device status server will call this endpoint whenever any device status related event occurs.
+              operationId: postNotification
+              requestBody:
+                required: true
+                content:
+                  application/cloudevents+json:
+                    schema:
+                      $ref: "#/components/schemas/CloudEvent"
+                    examples:
+                      ROAMING_STATUS:
```
akoshunyadi, line 190: Do we want to keep these old constants, since they are not used for the type?

code/API_definitions/device-status.yaml
```diff
@@ -496,44 +564,63 @@ components:
...
-          $ref: '#/components/schemas/RoamingEventType'
+          $ref: "#/components/schemas/Device"
+        type:
+          $ref: "#/components/schemas/RoamingEventType"
 
     RoamingEventType:
       type: string
       description: |
-        ROAMING_STATUS - Event triggered when the device switch from roaming ON to roaming OFF and conversely
+        roaming-status - Event triggered when the device switch from roaming ON to roaming OFF and conversely
```
akoshunyadi, line 584: I think it would be better to use SubscriptionType or SubscriptionEventType instead of RoamingEventType. 
- to better distinguish between subscription event type and notification event type
- as a preparation for other type of events

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
+        - type
+        - time
       properties:
-        eventType:
+        id:
           type: string
-        eventTime:
-          $ref: '#/components/schemas/EventTime'
-        eventId:
+          description: identifier of this event, that must be unique in the source context.
```
akoshunyadi, line 659: example missing

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
+        type:
           type: string
-          format: uuid
-        eventDetail:
+          description: "type of event as defined in each CAMARA API"
+          example: org.camaraproject.device-status.v0.roaming-status
+        specversion:
+          type: string
+          description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
+          example: 1.0
```
akoshunyadi, line 669: string examples are sometimes in quotes sometimes not, not consistent (this is not new with this PR, just noticed :-) )

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
         - type: object
           properties:
-            eventDetail:
-              $ref: '#/components/schemas/RoamingStatus'
+            data:
+              $ref: "#/components/schemas/RoamingStatus"
 
     EventRoamingOn:
-      description:  event structure for roaming on and roaming off change
+      description: event structure for roaming on and roaming off change
```
akoshunyadi, line 717: This is only for roaming on

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
-              enum:
-                - ROAMING_ON
-            eventDetail:
-              $ref: '#/components/schemas/RoamingChange'
+            data:
+              $ref: "#/components/schemas/RoamingChange"
 
     EventRoamingOff:
-      description:  event structure for roaming on and roaming off change
+      description: event structure for roaming on and roaming off change
```
akoshunyadi, line 726: This is only for roaming off

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
           properties:
-            eventDetail:
-              $ref: '#/components/schemas/SubscriptionEnds'
+            data:
+              $ref: "#/components/schemas/SubscriptionEnds"
 
     RoamingStatus:
       description: Event detail structure for ROAMING_STATUS event
-      type : object
+      type: object
```
akoshunyadi, line 754: There is no ROAMING_STATUS event anymore

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
-          $ref: '#/components/schemas/CountryCode'
+          $ref: "#/components/schemas/CountryCode"
         countryName:
-          $ref: '#/components/schemas/CountryName'
+          $ref: "#/components/schemas/CountryName"
 
     RoamingChange:
       description: Event detail structure for ROAMING_ON & ROAMING_OFF event
-      type : object
+      type: object
```
akoshunyadi, line 771: There are no ROAMING_ON and ROAMING_OFF events anymore

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +630,182 @@ components:
...
         countryName:
-          $ref: '#/components/schemas/CountryName'
+          $ref: "#/components/schemas/CountryName"
+        subscriptionId:
+          $ref: "#/components/schemas/SubscriptionId"
 
     SubscriptionEnds:
       description: Event detail structure for SUBSCRIPTION_ENDS event
-      type : object
+      type: object
```
akoshunyadi, line 797: There is no SUBSCRIPTION_ENDS event

code/API_definitions/device-status.yaml
bigludo7: I've changed the constant name to be closer to the type but I kept only the significant part;

bigludo7 (commented):
Took into account @akoshunyadi review comments.

code/API_definitions/device-status.yaml
```diff
@@ -17,11 +83,11 @@ security:
   - oAuth2ClientCredentials: []
   #  - BasicAuth: []
   #  - apiKey: []
-  - three_legged:
-      - device-status-roaming-read
-      - device-status-roaming-event
+  - threeLegged:
+      - device-status:roaming:read
+      - device-status:roaming-event-subscriptions:create
```
fernandopradocabrillo, line 88: How are we going to manage the scopes for the subscriptions? I can see that we have an scope for creating subscriptions but I think we are missing an additional scope to delete subscriptions too. And, in order to retrieve the available subscriptions, are we going to use the same scope as we use to read the roaming status?

```suggestion
      - device-status:roaming:read
      - device-status:roaming-event-subscriptions:read
      - device-status:roaming-event-subscriptions:create
      - device-status:roaming-event-subscriptions:delete
```

Whether we decide to use specific scopes or not, we should use the `security` property for each individual operation because right now, using the global property some operations don't have a correct scope. For example we have a create scope for a delete operation:

image

code/API_definitions/device-status.yaml
```diff
@@ -183,28 +253,28 @@ paths:
...
-          $ref: '#/components/responses/Generic403'
+          $ref: "#/components/responses/Generic403"
         500:
-          $ref: '#/components/responses/Generic500'
+          $ref: "#/components/responses/Generic500"
         503:
-          $ref: '#/components/responses/Generic503'
-  /event-subscriptions/{eventSubscriptionId}:
+          $ref: "#/components/responses/Generic503"
+  /subscriptions/{subscriptionId}:
```
sachinvodafone, line 267: It seems, we have missed security for this endpoint 
security:
        - threeLegged:

code/API_definitions/device-status.yaml
```diff
@@ -32,77 +98,33 @@ servers:
...
       responses:
         200:
           description: Contains information about current roaming status
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/RoamingStatusResponse'
+                $ref: "#/components/schemas/RoamingStatusResponse"
               examples:
                 No-Country-Code:
```
sachinvodafone, line 129: Is it example of "No-Country-Code" or "No-Country-Name"
Screenshot 2023-10-25 at 12 12 06

sachinvodafone:
In some responses, we are having missing "status": 
Screenshot 2023-10-25 at 12 29 30

sachinvodafone:
What kind of response can the client anticipate when they receive a subscription ID in a 202 response code (asynchronous process) and attempt to access /subscriptions/{subscriptionId} where process is still not completed  ?

1. 202 Accepted (Still Processing):
2. 404 Not Found (Subscription ID Not Found)
Screenshot 2023-10-25 at 12 43 57

bigludo7:
> What kind of response can the client anticipate when they receive a subscription ID in a 202 response code (asynchronous process) and attempt to access /subscriptions/{subscriptionId} where process is still not completed ?
> 
> 1. 202 Accepted (Still Processing):
> 2. 404 Not Found (Subscription ID Not Found)
> 
> Screenshot 2023-10-25 at 12 43 57

Good question and probably should tackled in a larger extend than this API. I will imagine 202 is a better option.

code/API_definitions/device-status.yaml
bigludo7: Fixed

bigludo7 (commented):
Updated following @sachinvodafone and @fernandopradocabrillo  review

sachinvodafone:
> > What kind of response can the client anticipate when they receive a subscription ID in a 202 response code (asynchronous process) and attempt to access /subscriptions/{subscriptionId} where process is still not completed ?
> > 
> > 1. 202 Accepted (Still Processing):
> > 2. 404 Not Found (Subscription ID Not Found)
> > 
> > Screenshot 2023-10-25 at 12 43 57
> 
> Good question and probably should tackled in a larger extend than this API. I will imagine 202 is a better option.

So in this scenario, we must have 202 in response code against "/subscriptions/{subscriptionId}"

sachinvodafone:
Typo mistake "0" instead of "500" error code without valid code and message in response example:

Screenshot 2023-10-27 at 11 04 07

bigludo7 (commented):
Fixed typo for error 500 (Thanks @sachinvodafone )

bigludo7 (commented):
Added 202 for GET /subscriptions/{subscriptionId} as suggested by @sachinvodafone

sachinvodafone (dismissed):

bigludo7:
Thanks @sachinvodafone for the approval;
@akoshunyadi and @fernandopradocabrillo  looking also for your approval or comments. Thanks

fernandopradocabrillo:
> > > What kind of response can the client anticipate when they receive a subscription ID in a 202 response code (asynchronous process) and attempt to access /subscriptions/{subscriptionId} where process is still not completed ?
> > > 
> > > 1. 202 Accepted (Still Processing):
> > > 2. 404 Not Found (Subscription ID Not Found)
> > > 
> > > Screenshot 2023-10-25 at 12 43 57
> > 
> > 
> > Good question and probably should tackled in a larger extend than this API. I will imagine 202 is a better option.
> 
> So in this scenario, we must have 202 in response code against "/subscriptions/{subscriptionId}"

I'm not really sure a 202 is the better option to return. If I create a subscription, receive a 202 and request the info before it is created, the subscription hasn't been created yet so I'd say it is better a 404.

I haven't seen a 202 in a GET, it is a bit weird. I think it adds complexity to the implementation because the response may vary if I make the call and receive a 202 and 5mins later I receive a 200 with the info, this does not happen with the POST/PUT operations since you only make the call once. And this will force the backend to have some way of tracking the process of creating the subscription to know what to return depending on the moment.

IMO I think it is more common to return a 404 and when the subscription is created, return the 200.

cc: @bigludo7 @sachinvodafone

code/API_definitions/device-status.yaml
```diff
@@ -131,50 +144,101 @@ paths:
...
-        - Device status event subscription
-      summary: 'Create a device status event subscription for a device'
+        - Device status subscription
+      summary: "Create a device status event subscription for a device"
       description: Create a device status event subscription for a device
-      operationId: createDevicStatusEventSubscription
+      operationId: createDeviceStatusSubscription
+      security:
+        - openId:
+            - device-status:roaming-subscriptions:create
```
akoshunyadi, line 156: It is a generic operation to create a subscription as the endpoint, description, operationId say, so I think also the scope shouldn't contain "roaming".

Similar for other subscription operations too.

bigludo7 (commented):
Remove 202 for GET following @fernandopradocabrillo  comment.
Rename scope Following @akoshunyadi comment

code/API_definitions/device-status.yaml
bigludo7: @akoshunyadi Yes make sense. Fixed.

bigludo7:
> > > > What kind of response can the client anticipate when they receive a subscription ID in a 202 response code (asynchronous process) and attempt to access /subscriptions/{subscriptionId} where process is still not completed ?
> > > > 
> > > > 1. 202 Accepted (Still Processing):
> > > > 2. 404 Not Found (Subscription ID Not Found)
> > > > 
> > > > Screenshot 2023-10-25 at 12 43 57
> > > 
> > > 
> > > Good question and probably should tackled in a larger extend than this API. I will imagine 202 is a better option.
> > 
> > 
> > So in this scenario, we must have 202 in response code against "/subscriptions/{subscriptionId}"
> 
> I'm not really sure a 202 is the best option to return. If I create a subscription, receive a 202 and request the info before it is created, the subscription hasn't been created yet so I'd say it is better a 404.
> 
> I haven't seen a 202 in a GET, it is a bit weird. I think it adds complexity to the implementation because the response may vary if I make the call and receive a 202 and 5mins later I receive a 200 with the info, this does not happen with the POST/PUT operations since you only make the call once. And this will force the backend to have some way of tracking the process of creating the subscription to know what to return depending on the moment.
> 
> IMO I think it is more common to return a 404 and when the subscription is created, return the 200.
> 
> cc: @bigludo7 @sachinvodafone

OK!
I guess need extra discussion for adding 202 for GET when we have potentially ASYNC creation. As this topic is larger to this API I have removed the 202 and adding this point here: https://github.com/camaraproject/Commonalities/issues/83
@sachinvodafone hopes it's ok for you.

code/API_definitions/device-status.yaml
```diff
@@ -543,170 +603,183 @@ components:
...
+      format: uri-reference
+      minLength: 1
+      description: "Identifies the context in which an event happened in the specific Provider Implementation."
+      example:
+        - https://github.com/cloudevents
+        - mailto:cncf-wg-serverless@lists.cncf.io
+        - urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        - cloudevents/spec/pull/123
+        - /sensors/tn-1234567/alerts
+        - 1-555-123-4567
```
rartych, line 673: Here we should give one `example` or multiple `examples` coherent with examples of events (at the end of the file)
It can be:
- Absolute URI like: `https://notificationSendServer12.supertelco.com` starting with **uri-scheme**
- Relative Reference like:  `/supertelco/notification/SendServer12`
- 
```suggestion
      example: https://notificationSendServer12.supertelco.com
```

code/API_definitions/device-status.yaml
rartych: Without quotes it is rendered by Swager as integer, just: 1

```suggestion
          example: '1.0'
```

code/API_definitions/device-status.yaml
```diff
@@ -780,52 +855,70 @@ components:
...
       value:
-        eventSubscriptionId: 123655
-        event:
-          eventType: ROAMING_OFF
-          eventTime: 2023-01-17T13:18:23.682Z
-          eventDetail:
-            device:
-              phoneNumber: 123456789
+        id: 123655
+        source: supertelco.notificationSendServer12
```
rartych, line 875: Changed to _Internet-wide unique URI with a DNS authority_:
```suggestion
        source: https://notificationSendServer12.supertelco.com
```

code/API_definitions/device-status.yaml
```diff
@@ -780,52 +855,70 @@ components:
...
       value:
-        eventSubscriptionId: 123655
-        event:
-          eventType: ROAMING_ON
-          eventTime: 2023-01-17T13:18:23.682Z
-          eventDetail:
-            device:
-              phoneNumber: 123456789
+        id: 123656
+        source: supertelco.notificationSendServer12
```
rartych, line 888: Changed to _Internet-wide unique URI with a DNS authority_:
```suggestion
        source: https://notificationSendServer12.supertelco.com
```

code/API_definitions/device-status.yaml
```diff
@@ -780,52 +855,70 @@ components:
...
       value:
-        eventSubscriptionId: 123655
-        event:
-          eventType: ROAMING_CHANGE_COUNTRY
-          eventTime: 2023-01-17T13:18:23.682Z
-          eventDetail:
-            device:
-              phoneNumber: 123456789
+        id: 123657
+        source: supertelco.notificationSendServer12
```
rartych, line 901: Changed to _Internet-wide unique URI with a DNS authority_:
```suggestion
        source: https://notificationSendServer12.supertelco.com
```

code/API_definitions/device-status.yaml
```diff
@@ -780,52 +855,70 @@ components:
...
       value:
-        eventSubscriptionId: 123655
-        event:
-          eventType: SUBSCRIPTION_ENDS
-          eventTime: 2023-01-17T13:18:23.682Z
-          eventDetail:
-            device:
-              phoneNumber: 123456789
+        id: 123658
+        source: supertelco.notificationSendServer12
```
rartych, line 915: Changed to _Internet-wide unique URI with a DNS authority_:
```suggestion
        source: https://notificationSendServer12.supertelco.com
```

code/API_definitions/device-status.yaml
```diff
@@ -1,7 +1,73 @@
...
+    It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely. 
+
+    ### Notification callback
+
+    This endpoint describes the event notification received on subscription listener side when the event occurred. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
+
+    **WARNING**: This callback endpoint must be exposed on the listener side as `POST /{$request.body#/webhook/notificationUrl}`
+      will be posted by the device status resource server to the webhook url provided by notification 
+      listener.
+
```
rartych, line 66: I think this section can be simplified:

```suggestion
    ### Notifications callback

    The  `notifications` callback describes the format of event notifications and expected responses to the messages sent when the event occurs. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.

    **WARNING**: This callback endpoint must be exposed and reachable on the listener side under 'notificationUrl` defined in the `webhook` attribute.

```

sachinvodafone (approved):

PR #87, opened 2023-12-01
Create release candidate 0.5.0-rc

akoshunyadi:
#### What type of PR is this?
* documentation
* subproject management

#### What this PR does / why we need it:

Create release candidate 0.5.0-rc

#### Which issue(s) this PR fixes:

Fixes #82 

#### Special notes for reviewers:

Should be merged after the PR merged for #84

PR #90, opened 2023-12-08
Make subscriptionDetail/device and CloudEvent/data mandatory

### LINKED ISSUE #84: Minor corrections in Subscription model before release
## Problem description
**`data` property should be required in callback request body**
In the operacion "postNotification", the POST /subscriptions callback, we have the property "data" as optional, but shouldn't it be required? Otherwise it exists the posibility of sending an empty notification. 
In the Sim Swap Subscriptions yaml we have it as required to avoid this issue.

**`subscriptionDetail.device` property should be required**
We have `device` as optional in all subscription endpoints, request body and responses. I think this one should be required too. For the /roaming and /connectivity operations it is required so the system knows what devices are you asking for. It should be the same for the notifications. If we leave it like it is right know a client can create a subscription and not specify for what device. The same for the the GET endpoints, right now, it is allowed to not provide information regarding the device the subscriptions belongs to.

akoshunyadi:
#### What type of PR is this?

* enhancement/feature

Make subscriptionDetail/device and CloudEvent/data mandatory

Fixes #84

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -652,6 +652,7 @@ components:
       type: object
       required:
         - type
+        - device
       properties:
         device:
           $ref: "#/components/schemas/Device"
@@ -752,6 +753,7 @@ components:
         - specversion
         - type
         - time
+        - data
       properties:
         id:
           type: string


fernandopradocabrillo (approved):
LGTM

bigludo7 (approved):
LGTM

sachinvodafone (approved):

bigludo7:
@akoshunyadi we can merge it ou you prefer to merge before #87 ?

PR #106, opened 2024-02-07
fix: move "countryCode" out of "device" in ROAMING_CHANGE_COUNTRY-example

maxl2287:
#### What type of PR is this?

Add one of the following kinds:
* correction

#### What this PR does / why we need it:
Moves the "countryCode" parameter out of the "device"-parameter.
Adds "countryName" inside of the corresponding example.

PR #109, opened 2024-02-07
fix: add missing "subscriptionId" in "RoamingStatus"-data for CloudEvent

maxl2287:
#### What type of PR is this?

* correction

In the examples of callback CloudEvents for roaming-status, there is the subscriptionId given, but not in the corresponding component.
This will be fixed with this PR.

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -99,7 +99,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.5.0
+  version: 0.6.0-wip
 externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
@@ -893,6 +893,8 @@ components:
           $ref: "#/components/schemas/CountryCode"
         countryName:
           $ref: "#/components/schemas/CountryName"
+        subscriptionId:
+          $ref: "#/components/schemas/SubscriptionId"
 
     RoamingChangeCountry:
       description: Event detail structure for org.camaraproject.device-status.v0.roaming-on event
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
Good Catch !

sachinvodafone:
Hi @maxl2287 , It appears that this pull request aims to address two issues. First, correcting the device object where the country code was incorrectly placed under the device object. Second, ensuring that the country name is represented as an array, not a string, in the callback event . As per my understanding, It does not involve adding or amending the subscription field, as indicated in the request description, am i correct ?

maxl2287:
@sachinvodafone you're right!
That fixes with the country-code came in caused by the previous PR: #106

bigludo7 (dismissed):
LGTM

akoshunyadi (dismissed):

fernandopradocabrillo:
Just one comment here. Shouldn't this be a fix? instead of going in the v0.6.0-wip, couldn't we include it in a v0.5.1?

@maxl2287 @bigludo7 @akoshunyadi

maxl2287:
> > Just one comment here. Shouldn't this be a fix? instead of going in the v0.6.0-wip, couldn't we include it in a v0.5.1?
> > @maxl2287 @bigludo7 @akoshunyadi
> 
> any thoughts? Once a team starts to implement the 0.5.0 version it's going to be obvious that there is a bug and I don't think that waiting to release a 0.6.0 is the best option. wdyt?
> 
> @maxl2287 @bigludo7 @akoshunyadi

I aggree.

@bigludo7 @akoshunyadi what about 0.5.1-wip ?

akoshunyadi:
The general idea in Camara is to make more -smaller- releases, so 0.6.0 should come quite soon, I would say. But I don't have any objection, if somebody needs a 0.5.1 even sooner.

bigludo7 (approved):
Good for me. Thanks

akoshunyadi:
I'll create an issue for 0.5.1, where we can discuss and define the content of it.

PR #129, opened 2024-03-13
fix: change INVALID_INPUT for INVALID_ARGUMENT according to guidelines

fernandopradocabrillo:
#### What type of PR is this?

* bug

Change old INVALID_INPUT 400 errors for INVALID_ARGUMENTS as stated in the commonalities guidelines

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -361,24 +361,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/SubscriptionInfo"
         "400":
-          description: Invalid input
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/ErrorInfo"
-              examples:
-                Generic400:
-                  summary: Schema validation failed
-                  value:
-                    code: INVALID_INPUT
-                    status: 400
-                    message: "Schema validation failed at  ..."
-                subscriptionIdRequired:
-                  summary: eventSubscription id is required
-                  value:
-                    code: INVALID_INPUT
-                    status: 400
-                    message: "Expected property is missing: subscriptionId"
+          $ref: '#/components/responses/SubscriptionIdRequired'
         "401":
           $ref: "#/components/responses/Generic401"
         "403":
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM

code/API_definitions/device-status.yaml
```diff
@@ -997,6 +963,25 @@ components:
...
+              value:
+                status: 400
+                code: INVALID_ARGUMENT
+                message: "Client specified an invalid argument, request body or query param"
+            subscriptionIdRequired:
+              summary: subscription id is required
+              value:
+                status: 400
+                code: INVALID_ARGUMENT
+                message: "Expected property is missing: subscriptionId"
```
akoshunyadi, line 984: 1. If the type is a special error, then the description could be also more concrete. 
2. This special 400 error shouldn't have a Generic400 as example

fernandopradocabrillo: I'm not completely getting the special error comment. Since in CAMARA we are using examples to include specific errors, I chose to create a different schema containing both examples. So this way we have two schemas, one containing only Generic 400 for the endpoints that doesn't have subscriptionId and another one containing both. 

Is there a better way to do it without adding it inline? or do you mean that it should only have the specific error because the Generic doesn't apply?

What description do you think would be more appropriate?

akoshunyadi: I thought that since we also have the Generic400 error, then this SubscriptionIdRequired is a specialized one, so it could have a 
- description like "SubscriptionId is missing" and
- it should only have one example, exactly this case when subscriptionid is missing:
  
          examples:
                status: 400
                code: INVALID_ARGUMENT
                message: "Expected property is missing: subscriptionId"

bigludo7 (approved):
LGTM

akoshunyadi (approved):
looks fine

fernandopradocabrillo:
Hi @akoshunyadi I think we can merge this one so we can include it in the changelog too before the release

Issue #110, opened 2024-02-08
Request for Addition of "last_updated_timestamp" Field in CAMARA DeviceStatus API


Sachinsiso:
**Problem Description**
The CAMARA version of the current DeviceStatus API lacks a crucial "ageOfInformation" or "last_updated_timestamp" field, resulting in inconsistent behavior across vendors. This absence poses challenges for application developers in determining the freshness of roaming status data.

**Expected Behavior**
Propose the addition of a "last_updated_timestamp" or similar field in the API response to indicate when the roaming status information was last updated. 

```
roaming: true
countryCode: 262
countryName: ["DE"]
last_updated_timestamp: “2018-04-05T17:31:00Z"
```

**Alternative Solution**
Explore and discuss any alternative approaches or fields that could address the issue. Consider other methods or data points that might serve the same purpose, assessing their feasibility and effectiveness.

**Additional Context**
NA

bigludo7:
Hi @Sachinsiso 
Do you expect to provide the latest date of the status change or the date of the last check ?

I mean, suppose I switched in roaming mode Jan 15 at 6:38 
Now  we are Feb 14 12:25

I make a request.
Suppose backend provide the information of current roaming situation on the fly
We agree that this lastRoamingCheckTime" should be Feb 14 12:25 ?

sachinvodafone:
Hi @bigludo7 , When verifying a device's roaming status through some central database such as HSS which keeps the correct status of device at all time, the recorded "lastRoamingCheckTime" will be established as "Feb 14 12:25."

In specific situations, the provider autonomously updates local roaming information upon receiving Diameter and SS7 signaling files. This update occurs both when a person initially enters the roaming state and whenever there is a signal change during roaming.

For instance, if an individual switched to roaming on "Jan 15 at 6:38," the system notes the "last_updated_timestamp" in its local repository at the time of this change. If the system checks the roaming status simultaneously, the "lastRoamingCheckTime" would be "Jan 15 at 6:38."  same as "last_updated_timestamp".

However, if their signal changes from 4G to 5G (or vice versa) on "Jan 20 at 7:00," the system receives updated Diameter/SS7 signal files, updating the "last_updated_timestamp" to "Jan 20 at 7:00." In this case, the "lastRoamingCheckTime" would then reflect "Jan 20 at 7:00." same as "last_updated_timestamp".

mhfoo:
[...]

PR #145, opened 2024-04-25
Addition of "lastChecked" Field

sachinvodafone:
What type of PR is this?

correction

What this PR does / why we need it:
The CAMARA version of the current DeviceStatus API lacks a crucial "lastChecked" field, resulting in inconsistent behavior across vendors. This absence poses challenges for application developers in determining the freshness of roaming status/connectivity data.

Which issue(s) this PR fixes:
Fixes https://github.com/camaraproject/DeviceStatus/issues/110

### DIFF
--- a/code/API_definitions/device-status.yaml
+++ b/code/API_definitions/device-status.yaml
@@ -45,7 +45,7 @@ info:
       - `CONNECTED_DATA`, if device is connected to the network via data usage
       - `NOT_CONNECTED`, if device is not connected to the network
 
-    * **lastChecked** : Indicate when the information provided was last confirmed to be correct.
+    * **LastStatusTime** : This property specifies the time when the status was last checked. Its inclusion in the response indicates that the information may not be current, while its absence suggests that the status information is fresh.
     # API Functionality
 
     The API exposes following capabilities:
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Hi @sachinvodafone 
Do you really think that we need to have this attribute **mandatory** in the response?
I was expecting to have this one as optional. Not a blocker for me.

code/API_definitions/device-status.yaml
```diff
@@ -466,24 +473,36 @@ components:
...
+        lastChecked:
+          $ref: "#/components/schemas/LastChecked"
         roaming:
           $ref: "#/components/schemas/ActiveRoaming"
         countryCode:
           $ref: "#/components/schemas/CountryCode"
         countryName:
           $ref: "#/components/schemas/CountryName"
 
+    LastChecked:
```
akoshunyadi, line 488: For me the name is not clear enough, what was checked? Maybe the name should follow the device-status pattern: LastStatusTime

fernandopradocabrillo:
> Hi @sachinvodafone Do you really think that we need to have this attribute **mandatory** in the response? I was expecting to have this one as optional. Not a blocker for me.

Hi, my understanding was the same. I also think that the property should be optional and, in addition to that, the descriptions should be updated accordingly.
By being optional we can define its behaviour as: if the property is not included in the response then the information is fresh. If it is included, it is not.

The desciption of the endpoint specifies: 
- Get the **current** roaming status and the country information
- Get the **current** connectivity status information

Now it may not be the current so the description should reflect that.

code/API_definitions/device-status.yaml
sachinvodafone: Since the field represents the time when the status was last checked, using "checked" instead of "status" would indeed be more appropriate and clearer in conveying the purpose of the field . However,  I'm open to updating the field name to either "LastCheckedTime" or "LastStatusTime". I'd like to hear feedback from others on which option the team prefers.

sachinvodafone:
> Hi @sachinvodafone Do you really think that we need to have this attribute **mandatory** in the response? I was expecting to have this one as optional. Not a blocker for me.

Noted and will update accordingly

code/API_definitions/device-status.yaml
fernandopradocabrillo: I also like "lastStatusTime". It is similar to what has been done in DeviceLocation with "lastLocationTime". https://github.com/camaraproject/DeviceLocation/blob/main/code/API_definitions/location-retrieval.yaml#L239

Issue #148, opened 2024-05-02
Extend ConnectivityStatus with `UNKNOWN`


maxl2287:
**Problem description**
When the service tries to call the network to get some information, but the network cannot deliver either information about DATA nor SMS, then we cannot respond with "NOT_CONNECTED", but we need to communicate to the consumer that the network was not able to deliver a current status of connection.

**Possible evolution**
Add a new connectivity-status `UNKNOWN`

sachinvodafone:
It seems, we have a similar issue against [issue-138](https://github.com/camaraproject/DeviceStatus/issues/138)

murthygorty:
hmm..mm @maxl2287 why wouldn't the API return a temporary error if it is unable to determine the connectivity status?
cc: @NoelWirzius @gmuratk

sachinvodafone:
There could be some potential use-cases for "Unknown" status:

- The network is partly down, so some services like voice calls work, but others like data or SMS don't work properly.
- When there is too much network traffic, some requests like data or SMS might be delayed or fail.
- The service provider's systems have problems, so they can't provide certain information.

bigludo7:
Tricky question...
I checked on Device location...

- In Location Retrieval if we are not able to locate the mobile within the required freshness we send back a 422
- In location Verification if we're not able to check the mobile location we send back a 200 with a "verificationResult": "UNKNOWN"

I tend to prefer 422 for our API as easier for my perspective to differentiate  "business" valuable response.

sachinvodafone:
Thanks @bigludo7 for more insight on this and I would also prefer 422 in this case where it explicitly indicates a failure to process the request due to some unknown network issue.

fernandopradocabrillo:
I also tend to agree with the 422. It should not be a successful response and we should not charge for something that is not client's fault.

maxl2287:
@jlurien @bigludo7 I guess we can transform this issue into: 
"Respond with HTTP-422 when the service is unable to provide information about the current connectivity".

Wdyt?

bigludo7:
> @jlurien @bigludo7 I guess we can transform this issue into: "Respond with HTTP-422 when the service is unable to provide information about the current connectivity".
> 
> Wdyt?

Yes !
Seems that we have a consensus to send back a 422 when we're not able to get the connectivity status.
We could probably make a PR for that and update this. What do you think @akoshunyadi ? are we still in time to do that for this release?

akoshunyadi:
[...]

PR #152, opened 2024-05-15
Separate endpoint yamls proposal for direct API

### LINKED ISSUE #125: Separate endpoints into different APIs
In this issue the topic of separating the different endpoints into separate APIs should be covered. 
The split makes sense to be aligned with the other Camara Subprojects like "Device Location". On the other hand this will also helps the productization of the APIs and a clear offering from the telco industries. 

Suggestion: 

- 1 API for Roaming + Roaming subscription 
- 1 API for Connectivity + Connectivity subscription

bigludo7:
#### What type of PR is this?

* enhancement/feature

Separate endpoint yaml for reachability & roaming.

Fixes #125 
issue for Device error identifier to be added.

Yaml for both roaming & reachability subscription provided once new subscription model merged in Commonalities
API name & endpoint to be validated by the team

#### Changelog input

```
 release-note
- Split yaml device status in 4 separate yaml for reachability-retrieval, roaming-retrieval, reachability-subscriptions, roaming-subscriptions.
- Align device error messages following commonalities device identifier error definition.

```

#### Additional documentation 

This section can be blank.

```
docs

```


bigludo7 (commented):
Fixed mega-Linter

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,397 @@
...
+  - url: "{apiRoot}/reachability-retrieval/v0"
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+tags:
+  - name: Device reachability status
+    description: Operations to get the current reachability status of a device
+paths:
+  /retrieve-reachability-status:
```
fernandopradocabrillo, line 60: @bigludo7 I see that for example in location-retrieval the WG opted for "/retrieval", in location-verification for "/verify", etc
If we have reachability-retrieval maybe we can go with "/retrieval" too? wdyt

bigludo7: Hello @fernandopradocabrillo 
Yes make sense !
For the API name, today is `reachability-retrieval` - do you think we need to shift to `reachability-status` and then have an endpoint `/retrieve` - This is perhaps more straighforward?

Same will happen for roaming.

maxl2287:
@bigludo7 @akoshunyadi 
Do we want to do here in this PR the split into 4 Yamls as well?

akoshunyadi:
@maxl2287 we can do the 2 subscription APIs in a 2nd PR to the splitting issue

bigludo7:
> @maxl2287 we can do the 2 subscription APIs in a 2nd PR to the splitting issue

Both work for me 
The most convenient for you Max.

maxl2287:
2nd PR is #161 for splitting `/subscriptions`

code/API_definitions/device-reachability-status.yaml
fernandopradocabrillo: sounds good to me, it looks clearer

akoshunyadi: /retrieve would be good. 
API name: I would prefer device-reachability-status, which contains the context 'device' for the case other reachability status API will come later.

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,397 @@
+openapi: 3.0.3
+info:
+  title: Device Reachability Status Retrieval
+  description: |
+    This API provides the customer with the ability to query device:
+      - Reachability status
```
akoshunyadi, line 6: This list suggests that other infos might come later, but this is a specialized API.

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,397 @@
...
+                    lastStatusTime: "2024-02-20T10:41:38.657Z"
+                    reachabilityStatus: NOT_CONNECTED
+        "400":
+          $ref: "#/components/responses/Generic400"
+        "401":
+          $ref: "#/components/responses/Generic401"
+        "403":
+          $ref: "#/components/responses/ReachabilityPermissionDenied403"
+        "404":
+          $ref: "#/components/responses/ReachabilityNotFound404"
```
fernandopradocabrillo, line 108: I can see that we have the generic NOT_FOUND in addition to the DeviceIdentifierNotFound. 
In which scenario does the API return an standard NOT_FOUND?

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,397 @@
...
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 401
+            code: "UNAUTHENTICATED"
+            message: "Request not authenticated due to missing, invalid, or expired credentials"
+    ReachabilityPermissionDenied403:
+      description: |
+        Client does not have sufficient permission.
+        In addition to regular scenario of `PERMISSION_DENIED`, other scenarios may exist:
+          - Phone number cannot be deducted from access token context.(`{"code": "NUMBER_VERIFICATION.INVALID_TOKEN_CONTEXT","message": "Phone number cannot be deducted from access token context"}`)
```
maxl2287, line 293: @bigludo7, is this a valid 403 error? 
Currently, we are not extracting or verifying a phone number from the access token. This situation seems more related to Number Verification rather than Device Status. 

Do you agree?

code/API_definitions/device-reachability-status.yaml
maxl2287: I agree with @fernandopradocabrillo 

Maybe we should more go the way of having a reachability Status `UNKNOWN` with HTTP - 200 than a HTTP-404 for Reachability not found.

see https://github.com/camaraproject/DeviceStatus/issues/148

@bigludo7 wdyt?

code/API_definitions/device-reachability-status.yaml
bigludo7: Good question :)
For me this API could be sent from the device itself in frontend flow and in this case we got the `phoneNumber` from the network (see ICM proposal [here](https://github.com/camaraproject/IdentityAndConsentManagement/pull/173)). I need to update the yaml to not have device mandatory anymore.

So I think we should keep this error to manage this case but of course shifting prefix from NUMBER_VERIFICATION to DEVICE_REACHEABILTY_STATUS.

WDYT?

code/API_definitions/device-reachability-status.yaml
bigludo7: @maxl2287 @fernandopradocabrillo I can remove the NOT_FOUND. No problem.

BTW we're still waiting outcomes for #148 which is a fair point. I will comment there.

code/API_definitions/device-reachability-status.yaml
fernandopradocabrillo: If we make the device not mandatory then, for me, it makes sense to keep the invalid token context error.

bigludo7: Yes @fernandopradocabrillo - This is my preferred option.

bigludo7 (commented):
Changed endpoint to /retrieve
Took suggestion on documentation
Device is not anymore mandatory
Add 422 for Network issue - Unable to provide reachability status

I think we're not on same page on the api name.

bigludo7:
Aligned device-roaming-status with the work done with device-reacheability-status.
Need review by the team (@akoshunyadi @fernandopradocabrillo @maxl2287 @sachinvodafone) :)

code/API_definitions/device-reachability-status.yaml
akoshunyadi: If we offer this possibility over the access token, then we should extend the documentation. Otherwise the user has no idea when he needs to add the device.
btw. the example still contains number-verification

bigludo7: Applied same fix for device-roaming-status api

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,397 @@
+openapi: 3.0.3
+info:
+  title: Device Reachability Status Retrieval
+  description: |
```
fernandopradocabrillo, line 4: we need to add the Auth section: https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#mandatory-template-for-infodescription-in-camara-api-specs

bigludo7: Applied

bigludo7:
Hi team
Seems that I've mixed myself for Device Reachability Status yaml (commit https://github.com/camaraproject/DeviceStatus/pull/152/commits/7e002c9b0a58c94eab6c1104e86a0365bc062ebf overrides https://github.com/camaraproject/DeviceStatus/pull/152/commits/b1b96c54cc21dbaa4f4c8bfa8806215f26d30e8c corrections).

I've fixed that (I hope so)  + added Authorization and authentication § in the doc as suggested by @fernandopradocabrillo for both yaml.

Sorry for the confusion.

maxl2287:
I aggree with @akoshunyadi.

bigludo7:
@akoshunyadi @maxl2287 added delete the old file in this PR

code/API_definitions/device-roaming-status.yaml
```diff
@@ -0,0 +1,432 @@
...
+          example:
+            status: 401
+            code: "UNAUTHENTICATED"
+            message: "Request not authenticated due to missing, invalid, or expired credentials"
+    RoamingPermissionDenied403:
+      description: |
+        Client does not have sufficient permission.
+        In addition to regular scenario of `PERMISSION_DENIED`, other scenarios may exist:
+          - Phone number cannot be deducted from access token context.(`{"code": "NUMBER_VERIFICATION.INVALID_TOKEN_CONTEXT","message": "Phone number cannot be deducted from access token context"}`)
+      headers:
```
akoshunyadi, line 324: Here it's a remaining of NUMBER_VERIFICATION, should be replaced

akoshunyadi: @bigludo7 could you please do this update, so we can approve and merge this PR? Thanks!

akoshunyadi (dismissed):
LGTM

bigludo7:
@sachinvodafone @fernandopradocabrillo may I ask you to review this one for blocking point as we target to merge this one soon. Thanks

sachinvodafone (dismissed):

fernandopradocabrillo (approved):
LGTM

code/API_definitions/device-reachability-status.yaml
```diff
@@ -0,0 +1,403 @@
...
+                  value:
+                    lastStatusTime: "2024-02-20T10:41:38.657Z"
+                    reachabilityStatus: CONNECTED_SMS
+                Connected-With-DATA:
+                  value:
+                    lastStatusTime: "2024-02-20T10:41:38.657Z"
+                    reachabilityStatus: CONNECTED_DATA
+                Not-Connected:
+                  value:
+                    lastStatusTime: "2024-02-20T10:41:38.657Z"
```
sachinvodafone, line 105: Do we have an example where network is connected to both SMS and Data  and we need to send a response accordingly like: {
  "lastStatusTime": "2024-02-20T10:41:38.657Z",
  "reachabilityStatus": "CONNECTED_SMS_DATA"
}

akoshunyadi (approved):

code/API_definitions/device-reachability-status.yaml
bigludo7: For me no as my understanding DATA means that SMS is also available.
@sachinvodafone I will merge this PR so probably we could continue the conversation in Discussions tab  as it does not impact this merging.

PR #158, opened 2024-06-07
Create connected-network-type.yaml

### LINKED ISSUE #143: Extend: Device Status Connectivity with Standard
If there is a valid Data connection established the API should also inform about the Standard with which the device is connected. (2G, 3g, 4g , 5gnsa, 5gsa) 

This is helping developers to get a better insight about the connection status.

gmuratk:
#### What type of PR is this?

* enhancement/feature

#143 requests this enhancement/feature as new API under Device Status.

Fixes #143 

Proposal doesn't include the subscriptions portion.

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -259,8 +259,6 @@ components:
       properties:
         device:
           $ref: "#/components/schemas/Device"
-      required:
-        - device
 
     ErrorInfo:
       type: object
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -173,8 +173,8 @@ components:
         - 5GSA
         - NON3GPP4G
         - NON3GPP5G
-        - NOT-CONNECTED
-        
+        - NOT_CONNECTED
+
     Device:
       description: |
         End-user equipment able to connect to a mobile network. Examples of devices include smartphones or IoT sensors/actuators.
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -16,13 +16,11 @@ info:
       At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
 
     * **Network Type** : Network Type is intended to provide insight to connected network's capabilities from standards perspective.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
-      - `2G`, if device is connected to the 2G network
-      - `3G`, if device is connected to the 3G network
-      - `4GLTE`, if device is connected to the 4G network
-      - `5GNSA`, if device's is connected to the 4G network but using 5G New Radio (NR) capability
-      - `5GSA`, if device is connected to the 5G network
-      - `NON3GPP4G`, if device is connected to the 4G network via WiFi network
-      - `NON3GPP5G`, if device's is connected to the 5G network via WiFi network
+      - `2G`, if device is connected to the 2G network technology
+      - `3G`, if device is connected to the 3G network technology
+      - `4G`, if device is connected to the 4G network technology
+      - `5G`, if device is connected to the 5G network technology
+      - `UNKNOWN` if connection [technology] can not be determined
 
 
     * **LastStatusTime** : This property specifies the time when the status was last updated. Its presence in the response indicates the freshness of the information, while its absence implies the information may be outdated or its freshness is uncertain.
@@ -90,22 +88,18 @@ paths:
               schema:
                 $ref: "#/components/schemas/ConnectedNetworkTypeResponse"
               examples:
-                Connected-to-4GLTE:
+                Connected-to-4G:
                   value:
                     lastStatusTime: "2024-02-20T10:41:38.657Z"
-                    connectedNetworkType: 4GLTE
-                Connected-To-5GNSA:
+                    connectedNetworkType: 4G
+                Connected-to-5G:
                   value:
                     lastStatusTime: "2024-02-20T10:41:38.657Z"
-                    connectedNetworkType: 5GNSA
-                Connected-To-Non3gpp4g:
+                    connectedNetworkType: 5G
+                Connected-to-Undetermined:
                   value:
                     lastStatusTime: "2024-02-20T10:41:38.657Z"
-                    connectedNetworkType: NON3GPP4G
-                Not-Connected:
-                  value:
-                    lastStatusTime: "2024-02-20T10:41:38.657Z"
-                    connectedNetworkType: NOT_CONNECTED
+                    connectedNetworkType: UNKNOWN
         "400":
           $ref: "#/components/responses/Generic400"
         "401":
@@ -156,24 +150,19 @@ components:
 
     ConnectedNetworkType:
       description: |
-        * NOT_CONNECTED: The device is not connected to network
+        * UNKNOWN: Used when connection [technology] can not be determined
         * 2G: 2nd Generation Mobile Communication Technology
         * 3G: 3rd Generation Mobile Communication Technology
-        * 4GLTE: 4th Generation Mobile Communication Technology aka LTE
-        * 5GNSA:4th Generation Mobile Communication Technology aka LTE with device connected 5G New Radio (NR)
-        * 5GSA: 5th Generation Mobile Communication Technology
-        * NON3GPP4G: Device connected via non-3GPP Radio Access Technology but connected to 4G Network (e.g. WiFi Calling)
-        * NON3GPP5G: Device connected via non-3GPP Radio Access Technology but connected to 5G Network (e.g. WiFi Calling)
+        * 4G: 4th Generation Mobile Communication Technology
+        * 5G: 5th Generation Mobile Communication Technology
+
       type: string
       enum:
         - 2G
         - 3G
-        - 4GLTE
-        - 5GNSA
-        - 5GSA
-        - NON3GPP4G
-        - NON3GPP5G
-        - NOT_CONNECTED
+        - 4G
+        - 5G
+        - UNKNOWN
 
     Device:
       description: |
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -50,19 +50,19 @@ externalDocs:
   url: https://github.com/camaraproject/
 
 servers:
-  - url: "{apiRoot}/network-type-retrieval/v0"
+  - url: "{apiRoot}/network-type-retrieval/vwip"
     variables:
       apiRoot:
         default: http://localhost:9091
         description: API root
 tags:
-  - name: Connected  Network Type
+  - name: Connected Network Type
     description: Operations to get the network type device is connected to
 paths:
   /retrieve-network-type:
     post:
       tags:
-        - Connected network type
+        - Connected Network Type
       summary: "Get the connected network type"
       description: Get the connected network type
       operationId: getConnectedNetworkType
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -15,10 +15,10 @@ info:
     * **Device**: A device refers to any physical entity that can connect to a network and participate in network communication.
       At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
 
-    * **Network Type** : Network Type is intended to provide insight to connected network's capabilities from standards perspective.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
-      - `2G`, if device is connected to the 2G network technology
-      - `3G`, if device is connected to the 3G network technology
-      - `4G`, if device is connected to the 4G network technology
+    * **Network Type** : Network Type is intended to provide insight into connected network's capabilities from standards perspective, and to reflect the mobile technology that would be displayed by the device to the end user where applicable.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
+      - `2G`, if device is connected to the 2G network technology (alternative indicators such as "G" or "E" may be displayed on the device)
+      - `3G`, if device is connected to the 3G network technology (alternative indicators such as "H" or "H+" may be displayed on the device)
+      - `4G`, if device is connected to the 4G network technology (alternative indicators such as "LTE" or "LTE+" may be displayed on the device)
       - `5G`, if device is connected to the 5G network technology
       - `UNKNOWN` if connection [technology] can not be determined
 
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -8,7 +8,7 @@ info:
 
     ## Connected Network Type
 
-    API consumer is able to inquire device's connected network type.
+    The API consumer is able to query the device's connected network type. The API provider can determine this in various ways. For example, the network type can be inferred from the specific radio cell to which the device is connected, as radio cells only support a single technology. It is important to note that the API does not query the device directly.
 
     # Relevant terms and definitions
 
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -33,7 +33,19 @@ info:
 
     The endpoint `POST /retrieve` allows to get connected Network Type.
 
+    # Identifying the device from the access token
 
+    This API requires the API consumer to identify a device as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
+
+    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
+
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
+
+    ## Error handling:
+    - If the subject cannot be identified from the access token and the optional `device` object is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
+
+    - If the subject can be identified from the access token and the optional `device` object is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
     ## Further info and support
 
     (FAQs will be added in a later version of the documentation)
@@ -314,11 +326,6 @@ components:
                 status: 403
                 code: "PERMISSION_DENIED"
                 message: "Client does not have sufficient permissions to perform this action"
-            InvalidTokenContext:
-              value:
-                status: 403
-                code: INVALID_TOKEN_CONTEXT
-                message: Invalid access token context
     ConnectedStatusNotFound404:
       description: Resource Not Found
       headers:
@@ -337,7 +344,7 @@ components:
             DeviceIdentifierNotFound:
               value:
                 status: 404
-                code: DEVICE_NOT_FOUND
+                code: IDENTIFIER_NOT_FOUND
                 message: Some identifier cannot be matched to a device
     Generic409:
       description: Conflict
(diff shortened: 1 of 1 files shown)


code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 401
+            code: "UNAUTHENTICATED"
+            message: "Request not authenticated due to missing, invalid, or expired credentials"
+    ConnectedNetworkTypePermissionDenied403:
+      description: |
+        Client does not have sufficient permission.
+        In addition to regular scenario of `PERMISSION_DENIED`, other scenarios may exist:
+          - Phone number cannot be deducted from access token context.(`{"code": "NUMBER_VERIFICATION.INVALID_TOKEN_CONTEXT","message": "Phone number cannot be deducted from access token context"}`)
```
maxl2287, line 312: Seems like a little bit of NumberVerification 👀 
```json
{
  "code": "NUMBER_VERIFICATION.INVALID_TOKEN_CONTEXT",
  "message": "Phone number cannot be deducted from access token context"
}
```

gmuratk: Thanks for the comment.  This is aligned with the other Device Status API endpoints.  If there is a change needed, I can apply here as well.

maxl2287: Hi @gmuratk,
which "other" DeviceStatus APIs are you referencing? 
I do not see any mention of verifying a number within the access token in those APIs.

The only instance I found related to this is in the [NumberVerification API](https://github.com/camaraproject/NumberVerification/blob/main/code/API_definitions/number_verification.yaml#L258-L263).

As we are not currently expecting to extract and verify phone numbers from the access token for the DeviceStatus API, do you think we can proceed with a general 403 response in such cases?

Wdyt?

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+              $ref: '#/components/headers/x-correlator'
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ConnectedNetworkTypeResponse"
+              examples:
+                Connected-to-4G:
+                  value:
+                    lastStatusTime: "2024-02-20T10:41:38.657Z"
+                    connectedNetworkType: 4G
```
maxl2287, line 96: 4G you have not used here inside your enum:

```yaml
  enum:
        - 2G
        - 3G
        - 4GLTE
        - 5GNSA
        - 5GSA
        - NON3GPP4G
        - NON3GPP5G
        - NOT-CONNECTED
```

Maybe you wanted to use `4GLTE` ?

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+      type: string
+      enum:
+        - 2G
+        - 3G
+        - 4GLTE
+        - 5GNSA
+        - 5GSA
+        - NON3GPP4G
+        - NON3GPP5G
+        - NOT-CONNECTED
```
maxl2287, line 176: Change this to `NOT_CONNECTED`

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+
```
maxl2287, line 177: Remove the trailing spaces in line 177

maxl2287 (changes requested):

code/API_definitions/connected-network-type.yaml
gmuratk: Hi @maxl2287, 
PR152, has the 2 files for separating the endpoints, and I used them as the basis for this yaml.  Error message you pointed out exist in those files as well. https://github.com/camaraproject/DeviceStatus/pull/152/files

maxl2287:
@camaraproject/device-status_codeowners I guess this can be moved together with #171 to the newly created https://github.com/camaraproject/DeviceQualityIndicator, do you agree?

code/API_definitions/connected-network-type.yaml
QaunainM: We are hearing of one use case more regularly "Companies want to check if a number is a VoIP number vs a nonVoIP number" , any chance we could get that as one of the options here?

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+        - connectedNetworkType
+      properties:
+        lastStatusTime:
+          $ref: "#/components/schemas/LastStatusTime"
+        connectedNetworkType:
+          $ref: "#/components/schemas/ConnectedNetworkType"
+
+    ConnectedNetworkType:
+      description: |
+        * NOT_CONNECTED: The device is not connected to network
```
eric-murray, line 159: I disagree with using `NOT_CONNECTED`. It should be `UNKOWN` (or `NOT_KNOWN`, or similar).

There are two reasons for this:
- There are various reasons why the API provider might not know the network type, and not being connected to the network at all is just one of them. Rather than have separate codes for all these reasons, and the API provider trying to work out what the reason is, better just to say "I don't know".
- The purpose of the API is to return the network type. Saying the device is not connected is providing information over and above this and, indeed, is information that is anyway provided by a different API (`device-reachability-status`). We should avoid having two APIs that do the same thing in different ways.

If there is an insistence that this API also returns information as to whether or not the device is connected to the network, then it should be an extension of `device-reachability-status` and not a new API.

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+        * 2G: 2nd Generation Mobile Communication Technology
+        * 3G: 3rd Generation Mobile Communication Technology
+        * 4GLTE: 4th Generation Mobile Communication Technology aka LTE
+        * 5GNSA:4th Generation Mobile Communication Technology aka LTE with device connected 5G New Radio (NR)
+        * 5GSA: 5th Generation Mobile Communication Technology
```
eric-murray, line 164: It will be hard enough trying t work out what radio technology the device is using, without having to also work out which core it is connected to. I would suggest a single `5G` option. What is the use case where the core technology is important?

Kevsy, line 164: Suggest to add 'NTN4G' and 'NTN5G' to the enum, as non-terrestrial (satellite) access is likely to incur increased propagation delay and reduced throughput vs. terrestrial 4G/5G.

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+        * NON3GPP4G: Device connected via non-3GPP Radio Access Technology but connected to 4G Network (e.g. WiFi Calling)
+        * NON3GPP5G: Device connected via non-3GPP Radio Access Technology but connected to 5G Network (e.g. WiFi Calling)
+      type: string
```
Kevsy, line 167:
```suggestion
        * NTN4G: Device connected over 4G via 3GPP non-terrestrial network 
        * NTN5G: Device connected over 5G via 3GPP non-terrestrial network 
      type: string
```

Kevsy: Suggestion above is to add non-terrestrial network (satellite) access, because it will incur increased propagation delay and reduced throughput vs. terrestrial 4G/5G

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+      enum:
+        - 2G
+        - 3G
+        - 4GLTE
+        - 5GNSA
+        - 5GSA
+        - NON3GPP4G
+        - NON3GPP5G
```
Kevsy, line 175:
```suggestion
        - NON3GPP5G
        - NTN4G
        - NTN5G
```

code/API_definitions/connected-network-type.yaml
eric-murray: > Suggestion above is to add non-terrestrial network (satellite) access

The problem with adding this option is that almost certainly nobody will support it. Why not? Because the device will almost certainly be roaming on such networks. And I don't see any API provider implementing a solution to determine what access technology a roaming device is using.

Non-terrestrial networks would be better detected using the roaming API, as they should be using a non-geographic PLMN.

eric-murray: > ... where satellite is filling coverage gaps to regular UEs

But you are still referring to a "direct to mobile" service using a 3GPP standard (4G or 5G)? Who is proposing to offer such a service using their normal terrestrial PLMN ID?

Kevsy: > Who is proposing to offer such a service using their normal terrestrial PLMN ID?

Too early to say, but with reports of dozens of markets considering satellite-coverage deployment, then it makes sense to me to cover the valid 'same PLMN ID' option in the API  (either as an enum value or with the optional modifier I mention above).

gmuratk:
Thank you for all the comments over the past months.  Since 'Draft' label is removed, we'll provide responses/edits soon.
cc: @murthygorty @RamTMO

code/API_definitions/connected-network-type.yaml
gmuratk: No additional update, so closing the conversation

code/API_definitions/connected-network-type.yaml
gmuratk: I will change to 'UNKNOWN'

code/API_definitions/connected-network-type.yaml
gmuratk: Updated to UNKNOWN as suggested

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,416 @@
...
+    ConnectedNetworkType:
+      description: |
+        * NOT_CONNECTED: The device is not connected to network
+        * 2G: 2nd Generation Mobile Communication Technology
+        * 3G: 3rd Generation Mobile Communication Technology
+        * 4GLTE: 4th Generation Mobile Communication Technology aka LTE
+        * 5GNSA:4th Generation Mobile Communication Technology aka LTE with device connected 5G New Radio (NR)
+        * 5GSA: 5th Generation Mobile Communication Technology
+        * NON3GPP4G: Device connected via non-3GPP Radio Access Technology but connected to 4G Network (e.g. WiFi Calling)
+        * NON3GPP5G: Device connected via non-3GPP Radio Access Technology but connected to 5G Network (e.g. WiFi Calling)
```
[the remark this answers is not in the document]
gmuratk: Removed options to simplify the list of connection technologies.  In future, list may be modified.

code/API_definitions/connected-network-type.yaml
gmuratk: Good discussion and thoughts, but it will be best to start with a simpler list of connection technology list.  As mentioned above, in future the list can be modified.

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: wip
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+
+servers:
+  - url: "{apiRoot}/network-type-retrieval/v0"
```
akoshunyadi, line 53: we need here vwip as a version "..../vwip"

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+tags:
+  - name: Connected  Network Type
```
bigludo7, line 59: Be careful small typo with line 65 Network - network

bigludo7 (commented):
One typo in the tag

code/API_definitions/connected-network-type.yaml
gmuratk: Thank you for noticing.  Even when looking it is hard to see!

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+    # API Functionality
+
+    The API exposes following capabilities:
+
+    ## Connected Network Type
+
+    The endpoint `POST /retrieve-network-type` allows to get connected Network Type.
+
+
```
akoshunyadi, line 36: If we allow device identifier in the access token, we need to add the template to the description: https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#appendix-a-normative-infodescription-template-for-when-user-identification-can-be-from-either-an-access-token-or-explicit-identifier

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+info:
+  title: Connected Network Type
+  description: |
+    This API provides the customer with the ability to query to which Mobile Communication Technology it is connected to.
+
+    # Introduction
+
+    ## Connected Network Type
+
+    API consumer is able to inquire device's connected network type.
```
eric-murray, line 11: Some additional text with an example implementation.
```suggestion
    The API consumer is able to query the device's connected network type. The API provider can determine this in various ways. For example, the network type can be inferred from the specific radio cell to which the device is connected, as radio cells only support a single technology. It is important to note that the API does not query the device directly.
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+    # Relevant terms and definitions
+
+    * **Device**: A device refers to any physical entity that can connect to a network and participate in network communication.
+      At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
+
+    * **Network Type** : Network Type is intended to provide insight to connected network's capabilities from standards perspective.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
```
eric-murray, line 18:
```suggestion
    * **Network Type** : Network Type is intended to provide insight into the connected network's capabilities from standards perspective, and to reflect the radio technology that would be displayed by the device to the end user where applicable.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+      - `2G`, if device is connected to the 2G network technology
+      - `3G`, if device is connected to the 3G network technology
```
eric-murray, line 20:
```suggestion
      - `3G`, if device is connected to the 3G network technology (may also be displayed on the device as "H" or "H+")
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+      - `4G`, if device is connected to the 4G network technology
```
eric-murray, line 21:
```suggestion
      - `4G`, if device is connected to the 4G network technology (may also be displayed on the device as "LTE" or "LTE+")
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+          examples:
+            NotFound:
+              value:
+                status: 404
+                code: NOT_FOUND
+                message: "The specified resource is not found"
+            DeviceIdentifierNotFound:
+              value:
+                status: 404
+                code: DEVICE_NOT_FOUND
```
eric-murray, line 340:
```suggestion
                code: IDENTIFIER_NOT_FOUND
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+          $ref: '#/components/headers/x-correlator'
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          examples:
+            UnsupportedDeviceIdentifiers:
+              value:
+                status: 422
+                code: UNSUPPORTED_DEVICE_IDENTIFIERS
```
eric-murray, line 368:
```suggestion
                code: UNSUPPORTED_IDENTIFIER
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+                message: "None of the provided device identifiers is supported by the implementation"
+            InconsistentDeviceIdentifiers:
+              value:
+                status: 422
+                code: DEVICE_IDENTIFIERS_MISMATCH
```
eric-murray, line 373:
```suggestion
                code: IDENTIFIER_MISMATCH
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+                message: Device identifiers mismatch
+            DeviceNotSupported:
+              value:
+                status: 422
+                code: DEVICE_NOT_APPLICABLE
```
eric-murray, line 378:
```suggestion
                code: SERVICE_NOT_APPLICABLE
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+                message: Service not applicable to the device
```
eric-murray, line 379:
```suggestion
                message: Service not applicable to the device
            UnnecessaryDeviceIdentifier:
              value:
                status: 422
                code: UNNECESSARY_IDENTIFIER
                message: The device is already identified by the access token
```

code/API_definitions/connected-network-type.yaml
gmuratk: @eric-murray, as you suggested in another change to this PR, identifying the connected technology based connected radio cell is only one option.  For some implementations it may not be possible to accurately identify the connected radio technology.  Would you be ok with the following change: ' ... from standards perspective, and to reflect the radio _**or core network**_ technology that would be displayed ...'

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,405 @@
...
+
+    API consumer is able to inquire device's connected network type.
+
+    # Relevant terms and definitions
+
+    * **Device**: A device refers to any physical entity that can connect to a network and participate in network communication.
+      At least one identifier for the device (user equipment) out of four options: IPv4 address, IPv6 address, Phone number, or Network Access Identifier assigned by the mobile network operator for the device.
+
+    * **Network Type** : Network Type is intended to provide insight to connected network's capabilities from standards perspective.  Actual network capabilities may differ based on implementation and MUST be checked with the connected network provider.
+      - `2G`, if device is connected to the 2G network technology
```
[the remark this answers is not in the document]
gmuratk: @eric-murray  Would it be better to include '_**Displayed connected technology indicator on device may differ from the types returned by this API**_' in the Network Type definition, and don't list out any specific values.  Such a list may be interpreted as all possible values.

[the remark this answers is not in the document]
eric-murray: I think the values I listed should cover most devices, but I agree it may well not be comprehensive. I often used to get asked "What does "E" or "H+" on my phone mean?", but not so much these days.

How about something like "alternative indicators such as "G" or "E" may be displayed on the device" ? So listing the common options, but saying these are just examples.

eric-murray:
@gmuratk
If you had time, you could add the modified error response schema in [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml) into the PR. But as this will be a WIP, we can add that later.

code/API_definitions/connected-network-type.yaml
gmuratk: @eric-murray  
I was thinking of the NSA cases where connected radio and core technology may not be the same.  In any case, generalizing is the way to go.

code/API_definitions/connected-network-type.yaml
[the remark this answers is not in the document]
gmuratk: Sounds good to me. Thx!

sachinvodafone (dismissed):

akoshunyadi (dismissed):

code/API_definitions/connected-network-type.yaml
```diff
@@ -0,0 +1,422 @@
...
+      format: ipv6
+      example: 2001:db8:85a3:8d3:1319:8a2e:370:7344
+
+    ConnectedNetworkTypeRequest:
+      type: object
+      properties:
+        device:
+          $ref: "#/components/schemas/Device"
+      required:
+        - device
```
eric-murray, line 263: Sorry - last minute review. `device` can't be required as a 3-legged token may be used. In that case, the request body would be an empty JSON object
```suggestion
```

akoshunyadi:
> @gmuratk If you had time, you could add the modified error response schema in [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml) into the PR. But as this will be a WIP, we can add that later.

I would suggest to merge this PR without the adapted error codes and work on those in a separate PR, for both APIs.

eric-murray:
> I would suggest to merge this PR without the adapted error codes and work on those in a separate PR, for both APIs.

That's fine, but we should remove that `device` is required before merging

eric-murray (approved):

akoshunyadi (approved):

sachinvodafone (approved):

bigludo7 (approved):
LGTM

murthygorty:
woohoo :-)

PR #170, opened 2024-06-18
Update the subscription models to align on CAMARA commonalities

### LINKED ISSUE #162: Align subscription APIs with new template
**Problem description**
The newly created subscription APIs (roaming and reachability) need to be updated to comply with the [subscription template](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml)

**Possible evolution**
Create a PR to adapt both APIs. No additional changes should be made.

**Alternative solution**

**Additional context**

maxl2287:
#### What type of PR is this?

* correction
* enhancement/feature

Update of the subscription-models based on: 
- https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#12-subscription-notification--event
- https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml

Fixes #162


akoshunyadi:
The base url should be with apiName, so 'device-' is missing..  (I can't comment there directly)

code/API_definitions/device-roaming-status-subscriptions.yaml
```diff
@@ -498,36 +704,9 @@ components:
         - org.camaraproject.roaming-status-subscriptions.v0.roaming-on
```
akoshunyadi, line 704: I guess we need here device- too, and then in the scopes too.

sachinvodafone (approved):

code/API_definitions/device-roaming-status-subscriptions.yaml
maxl2287: Aggree 🚀

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -25,24 +24,48 @@ info:
...
 
-      - ``org.camaraproject.reachability-status-subscriptions.v0.reachability-disconnected``: Event triggered when the device is not connected.
+      - ``org.camaraproject.device-reachability-status-subscriptions.v0.reachability-disconnected``: Event triggered when the device is not connected.
 
-    Note: Additionally to these list, ``org.camaraproject.reachability-status-subscriptions.v0.subscription-ends`` notification `type` is sent when the subscription ends.
+    Note: Additionally to these list, ``org.camaraproject.device-reachability-status-subscriptions.v0.subscription-ends`` notification `type` is sent when the subscription ends.
     This notification does not require dedicated subscription.
-    It is used when the subscription expire time (optionally set by the requester) has been reached or if the API server has to stop sending notification prematurely.
+
+    It is used when:
```
akoshunyadi, line 39: Rather "It is used in following cases" to emphasize that those lines are OR connected

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -25,24 +24,48 @@ info:
...
+
+    ### Authorization and authentication
+
+    [Camara Security and Interoperability Profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md) provides details on how a client requests an access token.
+
+    Which specific authorization flows are to be used will be determined during onboarding process, happening between the API Client and the Telco Operator exposing the API, taking into account the declared purpose for accessing the API, while also being subject to the prevailing legal framework dictated by local legislation.
+
+    It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
+
+    ### Authorization and authentication
```
akoshunyadi, line 62: This block seems to be duplicated

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -109,6 +132,12 @@ paths:
                     schema:
                       $ref: "#/components/schemas/CloudEvent"
                     examples:
+                      connectivity-data:
```
akoshunyadi, line 135: we might want to rename connectivity to reachability here too

code/API_definitions/device-roaming-status-subscriptions.yaml
```diff
@@ -38,13 +37,21 @@ info:
 
     Note: Additionally to these list, ``org.camaraproject.roaming-status-subscriptions.v0.subscription-ends`` notification `type` is sent when the subscription ends.
     This notification does not require dedicated subscription.
-    It is used when the subscription expire time (optionally set by the requester) has been reached or if the API server has to stop sending notification prematurely.
+
+    It is used when:
```
akoshunyadi, line 41: Rather "It is used in following cases" to emphasize that those lines are OR connected

code/API_definitions/device-roaming-status-subscriptions.yaml
```diff
@@ -38,13 +37,21 @@ info:
...
-    The `notifications` callback describes the format of event notifications and expected responses to the messages sent when the event occurs. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
+    This endpoint describes the event notification received on subscription listener side when the event occurred.
+    As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
 
-    **WARNING**: This callback endpoint must be exposed and reachable on the listener side under `notificationUrl` defined in the `webhook` attribute.
+    _**WARNING**: This callback endpoint must be exposed on the consumer side as `POST /{$request.body#/sink}`.
+      Developers may provide a callback URL on which notifications regarding reachability-status can be received from the service provider.
+      If an event occurs the application will send events to the provided webhook - `sink`._
 
     ## Further info and support
```
akoshunyadi, line 56: The block "Authentication..." seems to be missing here

bigludo7 (dismissed):
LGTM
Nice job Max ! thanks

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -25,29 +24,45 @@ info:
...
-    Following event ``type`` are managed for this API:
-      - ``org.camaraproject.reachability-status-subscriptions.v0.reachability-data``: Event triggered when the device is connected to the network for Data usage.
+    Following event ``types`` are managed for this API:
+      - ``org.camaraproject.device-reachability-status-subscriptions.v0.reachability-data``: Event triggered when the device is connected to the network for Data usage.
 
-      - ``org.camaraproject.reachability-status-subscriptions.v0.reachability-sms``: Event triggered when the device is connected to the network for SMS usage
+      - ``org.camaraproject.device-reachability-status-subscriptions.v0.reachability-sms``: Event triggered when the device is connected to the network for SMS usage
 
-      - ``org.camaraproject.reachability-status-subscriptions.v0.reachability-disconnected``: Event triggered when the device is not connected.
+      - ``org.camaraproject.device-reachability-status-subscriptions.v0.reachability-disconnected``: Event triggered when the device is not connected.
```
sachinvodafone, line 35: It stated "reachability disconnected" but didn't specify whether it pertains to "SMS" or "data," leading us to assume it applies to both. So just confirming if there is no use-case to send the notification separately for each.

sachinvodafone:
As per today discussion over subscription response, would apprecaite if we can include examples in specification that pointed in [discussion](https://github.com/camaraproject/DeviceStatus/discussions/168)

bigludo7:
@maxl2287 I have a second thought about making the device identifier optional in the subscription.
My point is the following: 
For the POST that's fine to not have this device identifier as we have it thru the access token.

But, when the developer check the subscription 10 days later, in the GET /subscriptions response better to have this device identifier present because without it impossible to know what device is targeted....

So... either 1/ we keep the device mandatory or 2/we make it optionnal for the POST request but mandatory in the GET response

I prefer 1/ but perhaps we can have other solution.

cc: @fernandopradocabrillo

code/API_definitions/device-reachability-status-subscriptions.yaml
maxl2287: As I understand it from the current specification, this `reachability-disconnected` - events occure when the device is not connected at all.
So you want to be informed when there is neither DATA nor SMS connection.

Please correct me if I am wrong @bigludo7

maxl2287:
> @maxl2287 I have a second thought about making the device identifier optional in the subscription. My point is the following: For the POST that's fine to not have this device identifier as we have it thru the access token.
> 
> But, when the developer check the subscription 10 days later, in the GET /subscriptions response better to have this device identifier present because without it impossible to know what device is targeted....
> 
> So... either 1/ we keep the device mandatory or 2/we make it optionnal for the POST request but mandatory in the GET response
> 
> I prefer 1/ but perhaps we can have other solution.
> 
> cc: @fernandopradocabrillo

Are you talking about https://github.com/camaraproject/DeviceStatus/issues/173 ?

maxl2287:
> As per today discussion over subscription response, would apprecaite if we can include examples in specification that pointed in [discussion](https://github.com/camaraproject/DeviceStatus/discussions/168)

@sachinvodafone I have added three examples for GET /subscriptions/{subscriptionId} pointing to specific `status` of the subscription.

sachinvodafone (dismissed):

code/API_definitions/device-reachability-status-subscriptions.yaml
sachinvodafone: I was reconfirming  if there is a separate event triggered when the device connects to the network for SMS and Data individually. Why do we only have one event triggered for disconnection?

sachinvodafone:
> > As per today discussion over subscription response, would apprecaite if we can include examples in specification that pointed in [discussion](https://github.com/camaraproject/DeviceStatus/discussions/168)
> 
> @sachinvodafone I have added three examples for GET /subscriptions/{subscriptionId} pointing to specific `status` of the subscription.

Thanks @maxl2287 , Can we get an example for "GET /subscriptions" where we provide only the subscriptionID for subscriptions with incomplete asynchronous calls and show the full response for the others (synchronous)?

maxl2287:
> Thanks @maxl2287 , Can we get an example for "GET /subscriptions" where we provide only the subscriptionID for subscriptions with incomplete asynchronous calls and show the full response for the others (synchronous)?

@sachinvodafone 
However, this will not align with the specified response for GET `/subscriptions`. 
The response for this endpoint includes a list of only having `Subscription` components. 

There should not be a mix of `Subscription` objects and `subscriptionId` values within the same list.

code/API_definitions/device-reachability-status-subscriptions.yaml
bigludo7: @maxl2287 Indeed as of now reachability-disconnected is when the mobile is not connected at all.

@sachinvodafone I got your point.
We have today 3 notification types: Data Connected, SMS connected and Disconnected. SMS connected mean that data is disconnected.
We have aligned this on the synchronous APIs response.

Alternatively we can define slightly distinct event as
- data-on 
- data-off
- sms-on
- sms-off

Max, Sachin wdyt?

bigludo7:
> > @maxl2287 I have a second thought about making the device identifier optional in the subscription. My point is the following: For the POST that's fine to not have this device identifier as we have it thru the access token.
> > But, when the developer check the subscription 10 days later, in the GET /subscriptions response better to have this device identifier present because without it impossible to know what device is targeted....
> > So... either 1/ we keep the device mandatory or 2/we make it optionnal for the POST request but mandatory in the GET response
> > I prefer 1/ but perhaps we can have other solution.
> > cc: @fernandopradocabrillo
> 
> @bigludo7 I aggree also on letting it mandatory here.

I suggest we do this way - let keep it mandatory. If in future we shift to optional this is not a breaking change.

sachinvodafone:
> > Thanks @maxl2287 , Can we get an example for "GET /subscriptions" where we provide only the subscriptionID for subscriptions with incomplete asynchronous calls and show the full response for the others (synchronous)?
> 
> @sachinvodafone However, this will not align with the specified response for GET `/subscriptions`. The response for this endpoint includes a list of only having `Subscription` components.
> 
> There should not be a mix of `Subscription` objects and `subscriptionId` values within the same list.

@maxl2287 I understand your point that it will only include those subscriptions that have already been created. Thank you.

maxl2287:
> > > Thanks @maxl2287 , Can we get an example for "GET /subscriptions" where we provide only the subscriptionID for subscriptions with incomplete asynchronous calls and show the full response for the others (synchronous)?
> > 
> > 
> > @sachinvodafone However, this will not align with the specified response for GET `/subscriptions`. The response for this endpoint includes a list of only having `Subscription` components.
> > There should not be a mix of `Subscription` objects and `subscriptionId` values within the same list.
> 
> @maxl2287 I understand your point that it will only include those subscriptions that have already been created. Thank you.

@sachinvodafone yes, but to be more precise: 
The list includes every subscription, which was created by the user, even those which are still in the creation process (see status ACTIVATION_REQUESTED).

sachinvodafone:
> > > > Thanks @maxl2287 , Can we get an example for "GET /subscriptions" where we provide only the subscriptionID for subscriptions with incomplete asynchronous calls and show the full response for the others (synchronous)?
> > > 
> > > 
> > > @sachinvodafone However, this will not align with the specified response for GET `/subscriptions`. The response for this endpoint includes a list of only having `Subscription` components.
> > > There should not be a mix of `Subscription` objects and `subscriptionId` values within the same list.
> > 
> > 
> > @maxl2287 I understand your point that it will only include those subscriptions that have already been created. Thank you.
> 
> @sachinvodafone yes, but to be more precise: The list includes every subscription, which was created by the user, even those which are still in the creation process (see status ACTIVATION_REQUESTED).

Suppose if we have only one subscription and that is asynchronous which still not completed, in this case ,we will get "ACTIVATION_REQUESTED"  if user try for "GET /Subscriptions" . If Yes, then I am having another question.

maxl2287:
> Suppose if we have only one subscription and that is asynchronous which still not completed, in this case ,we will get "ACTIVATION_REQUESTED" if user try for "GET /Subscriptions" . If Yes, then I am having another question.

You will receive something like:

```json
[
  {
    "sink": "https://endpoint.example.com/sink",
    "sinkCredential": {},
    "types": [
      "string"
    ],
    "config": {
      "subscriptionDetail": {
        "device": {
          "phoneNumber": "+123456789",
          "networkAccessIdentifier": "123456789@domain.com",
          "ipv4Address": {
            "publicAddress": "84.125.93.10",
            "publicPort": 59765
          },
          "ipv6Address": "2001:db8:85a3:8d3:1319:8a2e:370:7344"
        }
      },
      "subscriptionExpireTime": "2023-01-17T13:18:23.682Z",
      "subscriptionMaxEvents": 5,
      "initialEvent": true
    },
    "id": "550e8400-e29b-41d4-a716-446655440000",
    "startsAt": "2024-07-04T11:44:36.664Z",
    "expiresAt": "2024-07-04T11:44:36.664Z",
    "status": "ACTIVATION_REQUESTED"
  }
]
```

sachinvodafone:
> > Suppose if we have only one subscription and that is asynchronous which still not completed, in this case ,we will get "ACTIVATION_REQUESTED" if user try for "GET /Subscriptions" . If Yes, then I am having another question.
> 
> You will receive something like:
> 
> ```json
> [
>   {
>     "sink": "https://endpoint.example.com/sink",
>     "sinkCredential": {},
>     "types": [
>       "string"
>     ],
>     "config": {
>       "subscriptionDetail": {
>         "device": {
>           "phoneNumber": "+123456789",
>           "networkAccessIdentifier": "123456789@domain.com",
>           "ipv4Address": {
>             "publicAddress": "84.125.93.10",
>             "publicPort": 59765
>           },
>           "ipv6Address": "2001:db8:85a3:8d3:1319:8a2e:370:7344"
>         }
>       },
>       "subscriptionExpireTime": "2023-01-17T13:18:23.682Z",
>       "subscriptionMaxEvents": 5,
>       "initialEvent": true
>     },
>     "id": "550e8400-e29b-41d4-a716-446655440000",
>     "startsAt": "2024-07-04T11:44:36.664Z",
>     "expiresAt": "2024-07-04T11:44:36.664Z",
>     "status": "ACTIVATION_REQUESTED"
>   }
> ]
> ```

I have updated the scenario in discussion board [here](https://github.com/camaraproject/DeviceStatus/discussions/168)

maxl2287:
@bigludo7 can we do here a final review?
I just have added some examples based on @sachinvodafone's feedback.

bigludo7 (dismissed):
LGTM.
Thanks

bigludo7:
> @bigludo7 can we do here a final review? I just have added some examples based on @sachinvodafone's feedback.

Thanks @maxl2287 - Look good for me. Let's get final review from @sachinvodafone and @akoshunyadi and we're good to go.

akoshunyadi (commented):
The server url seems to miss the device- prefix for roaming

akoshunyadi (dismissed):

fernandopradocabrillo (approved):
LGTM

bigludo7 (dismissed):
LGTM

bigludo7:
@sachinvodafone it is ok for you now? Looking for you thumb up to merge :)
Thanks

code/API_definitions/device-reachability-status-subscriptions.yaml
maxl2287: @bigludo7 can we either create a discussion on that topic or an issue?
Maybe a better place to discuss :)

bigludo7: @maxl2287  Yes a discussion is a good idea. Do not block the PR for that. I copy/past las comment in a discussion.

bigludo7 (approved):
LGTM

akoshunyadi (approved):

PR #179, opened 2024-07-03
Switch device structure to optional

### LINKED ISSUE #173: Make Device attribute optional in device-roaming-status & device-reachability-status
**Problem description**
As discussed in Commonalities issue#[171](https://github.com/camaraproject/Commonalities/issues/171) & [PR233 ](url)we need to shift the `device` object to optional. 

**Expected behavior**
Apply the mechanism to rely on the access_token (not providing the device object in the API request) for 3-legged access scenarios. This would fix interoperability in all these scenarios, which is a huge improvement. We can define the device identifier as optional in the API specification and, if not provided, simply refer to access_tokehttps://github.com/camaraproject/Commonalities/pull/233n. The developer would simply not provide any further device information, which is simpler. The operator does not need to check that the device identifier actually matches the access_token provided and could simply rely on the access_token itself.

**Alternative solution**

**Additional context**

bigludo7:
#### What type of PR is this?

* enhancement/feature

Shift device structure cardinality to optional has the device identifier could be obtained from the 3-legs process.

Fixes #173 

#### Changelog input

```
 release-note
- Device structure is now optional has the device identifier could be obtained from the 3-legs process.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/device-reachability-status.yaml
+++ b/code/API_definitions/device-reachability-status.yaml
@@ -245,8 +245,6 @@ components:
       properties:
         device:
           $ref: "#/components/schemas/Device"
-      required:
-        - device
 
     ErrorInfo:
       type: object
--- a/code/API_definitions/device-roaming-status.yaml
+++ b/code/API_definitions/device-roaming-status.yaml
@@ -269,8 +269,6 @@ components:
       properties:
         device:
           $ref: "#/components/schemas/Device"
-      required:
-        - device
 
     ErrorInfo:
       type: object


bigludo7:
@sachinvodafone @akoshunyadi @fernandopradocabrillo Gentle reminder for this one as is it a simple one.
As we're close to the finish line and there is few other topic to adress for these API like error code.
Thanks

sachinvodafone (approved):

hdamker (approved):

PR #191, opened 2024-07-15
Make device attribute optional for subscription requests

### LINKED ISSUE #189: Make Device attribute optional for roaming and reachability subscriptions
**Problem description**
As stated in the [Camara design guideline](https://github.com/camaraproject/Commonalities/blob/r0.4.0-rc.1/documentation/API-design-guidelines.md) when using 3-legged access tokens and the device can be uniquely identified by the token, the Device identifier in the API request can be omitted.

**Possible evolution**
Make the attribute Device optional in the subscription creation request for both device-roaming-status-subscriptions and device-reachability-status-subscriptions.

**Alternative solution**

**Additional context**

akoshunyadi:
#### What type of PR is this?

* enhancement/feature

The device attribute in the subscription request has been made optional.

Fixes #189 

#### Changelog input

```
Make device attribute optional for subscription requests

```

### DIFF
--- a/code/API_definitions/device-reachability-status-subscriptions.yaml
+++ b/code/API_definitions/device-reachability-status-subscriptions.yaml
@@ -506,9 +506,7 @@ components:
       properties:
         device:
           $ref: "#/components/schemas/Device"
-      required:
-        - device
-
+  
     Subscription:
       description: Represents a event-type subscription.
       type: object
--- a/code/API_definitions/device-roaming-status-subscriptions.yaml
+++ b/code/API_definitions/device-roaming-status-subscriptions.yaml
@@ -510,8 +510,6 @@ components:
       properties:
         device:
           $ref: "#/components/schemas/Device"
-      required:
-        - device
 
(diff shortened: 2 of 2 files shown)


bigludo7:
Hi @akoshunyadi 
I'm a bit uncomfortable on this one as we must:

- in the notification send back the device identifier
- in the Subscription resource itself have the device identifier in the GET answer (or in POST response)

If we make it optional in the request, it means that the API server will have to get it from the access token and then store it in the subscription resource.
I tend to think that keeping it mandatory makes the thing easier - wdyt?

akoshunyadi:
@bigludo7 yes, that's a good point! I don't think that we should expose the phone-number from the access token either.

fernandopradocabrillo:
@akoshunyadi I agree with @bigludo7 on this one, I also prefer to have it required

akoshunyadi:
As decided in the last sub-project status call, the PR can be closed, see previous comments.

PR #208, opened 2024-09-16
Add missing `protocol`-components for device-roaming-status-subscriptions

### LINKED ISSUE #207: [Device-Roaming-Subscription]: Protocol components are missing in specification
**Describe the bug**
The components described in the `descriminator` for the `prototocol`-property are missing.
This causes problems with our code-generator and makes it unpossible to create the expected property-models.

See:
```yaml
      discriminator:
        propertyName: protocol
        mapping:
          HTTP: "#/components/schemas/HTTPSubscriptionRequest"
          MQTT3: "#/components/schemas/MQTTSubscriptionRequest"
          MQTT5: "#/components/schemas/MQTTSubscriptionRequest"
          AMQP: "#/components/schemas/AMQPSubscriptionRequest"
          NATS: "#/components/schemas/NATSSubscriptionRequest"
          KAFKA: "#/components/schemas/ApacheKafkaSubscriptionRequest"
```
Tested with [r1.2 - device-roaming-status-subscriptions.yaml](https://github.com/camaraproject/DeviceStatus/blob/r1.2/code/API_definitions/device-roaming-status-subscriptions.yaml)
**Expected behavior**
The components should be there, otherwise the yaml is invalid.

maxl2287:
#### What type of PR is this?

* bug

This PR adds the missing components, which were referenced in the `protocol`-discriminator

Fixes #207

### DIFF
--- a/code/API_definitions/device-reachability-status-subscriptions.yaml
+++ b/code/API_definitions/device-reachability-status-subscriptions.yaml
@@ -68,7 +68,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.6.0
+  version: wip
   x-camara-commonalities: 0.4.0
 externalDocs:
   description: Product documentation at CAMARA
--- a/code/API_definitions/device-reachability-status.yaml
+++ b/code/API_definitions/device-reachability-status.yaml
@@ -69,7 +69,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.6.0
+  version: wip
   x-camara-commonalities: 0.4.0
 externalDocs:
   description: Product documentation at CAMARA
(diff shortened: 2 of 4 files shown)


bigludo7 (approved):
LGTM

sachinvodafone (approved):

akoshunyadi (approved):

akoshunyadi (changes requested):
Sorry, just forgot about it. The first change after 0.6.0 should set the version back to wip.

bigludo7 (approved):
LGTM

akoshunyadi:
@maxl2287 now you set the version for all of the APIs to wip. I think generally it is not necessary, because it is possible that an API remains unchanged in the upcoming version, so only the changed API should be set to wip and later for the release to the new version. Of course we could set the previous version again, if there was no change... Anyway, let's keep it for now.

akoshunyadi (approved):

eric-murray (approved):

Issue #222, opened 2024-10-30
Enhance API documentation


akoshunyadi:
**Problem description**
From a review of @tanjadegroot :

For the novice user of this API (like me) the description is not sufficient: when do I subscribe to roaming status as opposed to roaming_on/off ? do I need to subscribe in addition to the countrychange event ? or is the roaming status event sent in case of country change ? Maybe to take up for the next release.
would be good to have a bit more explanation beyond just the list of events.

**Expected action**
Review and enhance the API documentation for all 4 APIs

**Additional context**

sachinvodafone:
My understanding on 'roaming-change-country" is as follows: 



| Use Case                                                          | Required Subscriptions                     | Note                                                                                                                                                                                      |
|-------------------------------------------------------------------|-------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|  |                          |                                                                                                                                                                                                                                                                                                                                                                                 |
| Notify me whenever the device changes countries while roaming     | `roaming-change-country`                   | - If you want to track the country changes while the device remains in a roaming state. For example, if a device moves from France to Germany while roaming, you’ll get notified.           |
|                                                                   |                                           | - The subscription will not end automatically when a device returns to its home country. The device subscription will remain active until:                                                 |
|                                                                   |                                           |           - You explicitly unsubscribe.                                                                                                             |
|                                                                   |                                           |           - The subscription expires (if there’s an expiration mechanism in place).                                                                                                 |
|                                                                   |                                           |   - However, no notifications will be sent from `roaming-change-country` unless the device re-enters a roaming state and subsequently changes countries.                                     |
| Notify me of any roaming status change and country changes        | `roaming-status` + `roaming-change-country` | You need to subscribe to 2 events in this scenario.                                                                                                                                       |

sachinvodafone:
> #218

In this scenario, an event will be generated only if the device is in roaming; otherwise, no event will be triggered.

akoshunyadi:
That's a good question, because if I go from not-roaming to roaming I won't receive a roaming-change-country, right?
But if subscribe in this situation to the event with initialEvent = true, then I should receive it?
My general understanding is, that the caller should receive an event immediately on initialEvent = true, when he would have received the same event by subscribing earlier to the event.  
Now in this situation we probably don't have the information whether we came from the home country or another roaming country.

gmuratk:
[...]

Issue #212, opened 2024-09-27
Support of Multi-SIM lines in DeviceStatus APIs


jlurien:
**Problem description**
As introduced in https://github.com/camaraproject/Commonalities/issues/302, for Multi-SIM services. providing a phoneNumber as input does not identify a single device uniquely.

Current DeviceStatus specs do not consider or give guidelines about how to behave in scenarios when the input is only a phoneNumber of a Multi-SIM service and specific device (SIM) cannot be inferred from the access token or other device identifier.
 
**Possible evolution**
Discuss the problem and alternatives. There may be impact in other tracks, such as IC&M. We may need also to discuss implications in the short term when dealing with this situation in current versions.

In the case of DeviceStatus each API has each own considerations, as each SIM may have a different status:

* For device-roaming-status, a user may have left one device linked to the multiSIM phone number at home while is travelling abroad with another one. Should API return roaming=True in that case? A corner case would be that several devices were roaming in different countries, and in that case a single countryCode/countryName couldn't be returned.
* For device-reachability-status, several SIMs may have different reachabilityStatus. Should API return a consolidated response taking into account if any of the SIMs is connected?
* For the subscription APIs, should an event be triggered if any of the SIMs changes its status? Should each device be identified uniquely in the event?

**Alternative solution**
- Relying on 3-legged tokens as much as possible would minimize the problem, but currently we cannot assure that they always identify a single device instead of a phone number. 
- Finally we can handle the situation with errors, such as 422 UNIDENTIFIABLE_DEVICE, or another dedicated code, informing clients to provide an alternative identifier that can identify the specific device. But this should be limited to corner situations as it is not dev friendly and in most cases developers do not know if a phone number belongs to a multi-SIM service.

(this discussion is similar to the one being held in https://github.com/camaraproject/DeviceLocation/issues/257)

eric-murray:
@jlurien

Discussing this internally for device-reachability-status, the conclusion was:
- if the MSISDN is the primary MSISDN, then it is the reachability status of the multi-SIM group that is required (so if any device is reachable, the assumption is that the customer themselves is reachable)
- if the MSISDN is a secondary MSISDN (i.e. the MSISDN of one of the secondary devices), then it is the reachability status of that device that is of interest

We can implement this using the existing specification based on the `Device` object, but using different logic dependent on whether the `phoneNumber` is a primary or secondary MSISDN.

[...]

PR #228, opened 2024-12-04
Update info.description

eric-murray:
#### Purpose:
* documentation

Updates `info.description` with:
- New [Commonalities template](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#appendix-a-normative-infodescription-template-for-when-user-identification-can-be-from-either-an-access-token-or-explicit-identifier) for User identification
- Add section on multi-SIM scenario handling

Fixes #None
But helps with #212 and #222

Will need to revisit once multi-SIM scenario handling is standardised

#### Changelog input

```
 release-note
 - Update info.description section for user identification and multi-SIM scenario handling
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/device-reachability-status.yaml
+++ b/code/API_definitions/device-reachability-status.yaml
@@ -53,7 +53,10 @@ info:
 
     # Multi-SIM scenario handling
 
-    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device for which the reachability status should be returned from that phone number. If the phone number is used as the device identifier when querying in a multi-SIM scenario, the API may respond with an error, return a combined status for the multi-SIM group as a whole, or return the status for a single device in the multi-SIM group which may not be the intended device.
+    In multi-SIM scenarios where more than one mobile device is associated with a phone number (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device for which the reachability status should be returned from that phone number. If the phone number is used as the device identifier when querying in a multi-SIM scenario, the API may:
+    - respond with an error, or
+    - return a combined status for the multi-SIM group as a whole, or
+    - return the status for a single device in the multi-SIM group, which may not be the intended device.
 
     Possible solutions in such a scenario include:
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
Look good for me. Indeed the multi-SIM could be updated later.

sachinvodafone (approved):

eric-murray:
I updated all YAMLs with text on multi-SIM scenario handling, and also the "identifying the device from the token" text for the device roaming API.

I had a question on identifying the device for the subscription APIs. Currently `device` is a required parameter of the Subscription Request. So does this mean that subscriptions can only be created with a 2-legged token? Or should `device` actually be optional?

eric-murray:
@camaraproject/device-status_maintainers 
`info.description` now updated for all APIs, and ready for review

@bigludo7
My question in the [comment](https://github.com/camaraproject/DeviceStatus/pull/228#issuecomment-2550975786) above is still open. Is it correct that a device identifier is required to create a subscription? If so, what are the rules when a 3-legged token is used to create the subscription?

akoshunyadi (approved):

bigludo7:
Hi @eric-murray 

> Is it correct that a device identifier is required to create a subscription?

At least yes for the subscriptions for device status or for geofencing.

> If so, what are the rules when a 3-legged token is used to create the subscription?

We have specified in the design guideline that: "If 3-legged access token is used, the POST and GET responses must not provide any device identifier.". Which mean that, like for non subscription API, we did not accept an identifier in the subscription body when we find the identifier from the token.

in the [event-subscription-template.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml) you find the 422 UNNECESSARY_IDENTIFIE error specified.

eric-murray:
@bigludo7
But if an explicit device identifier is always required, does that not prevent a 3-legged token from being used to create a subscription (because that will trigger the `422 UNNECESSARY_IDENTIFIER` error)?

In any case, I raised the issue explicitly in Issue #241. If it is correct that an explicit device identifier is always required, I'd like to understand the 3-legged token scenario for the documentation.

bigludo7 (approved):
LGTM

PR #232, opened 2024-12-05
Update device error model

### LINKED ISSUE #230: Align Device Errors with last Commonalities agreement for Spring25
**Problem description**

Current Error model is aligned with Commonalities v0.4 from Fall24 meta.

**Possible evolution**

Update error model to comply with new defined status codes and errors.
Sources:
- https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml
- https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#62-error-responses---device-objectphone-number

fernandopradocabrillo:
#### What type of PR is this?

* correction
* enhancement/feature
* documentation

Align error model with Commonalities for meta Spring25

Fixes #230 

In the subscriptions APIs there is one error that I'm not sure if we have to update or not.

```
TokenMismatch:
   value:
    status: 403
    code: SUBSCRIPTION_MISMATCH
    message: Inconsistent access token for requested events subscription
```

It looks like it is the equivalent to `403 INVALID_TOKEN_CONTEXT`. It may make sense to keep it since the `device` object is required when creating the subscription, regardless of the token type. I tend to think we should keep it.


bigludo7:
@fernandopradocabrillo Agree with you to keep it but in order to be explicit the code should not be: `INVALID_TOKEN_CONTEXT` ?

bigludo7:
Hi @fernandopradocabrillo 
Additional thinking:
- The callback must have 429 in device-roaming-status-subscriptions.yaml (in lines 140/155)

Following points are open:
- Do we keep 500, 501, 502, 503 & 504 in the yamls ? they are not anymore mandatory to be defined
- 429 is not mandatory for the GET /subscriptions/{subscriptionId} & GET /subscriptions. 
- same for DELETE

fernandopradocabrillo:
> SUBSCRIPTION_MISMATCH

I'm not sure, I see that in geofencing it is also SUBSCRIPTION_MISMATCH: https://github.com/camaraproject/DeviceLocation/blob/main/code/API_definitions/geofencing-subscriptions.yaml#L1166

do you know if there is any conversation right now in commonalities regarding the errors in subscriptions?

fernandopradocabrillo:
> @fernandopradocabrillo would it be possible to cover the 2 new APIs here as well? 🙂 (connected-network-type-subscriptions is still to be merged)

and I won't charge extra

fernandopradocabrillo:
@camaraproject/device-status_maintainers shouldn't the POST /subscription endpoint also have the 404 IDENTIFIER_NOT_FOUND error?

fernandopradocabrillo:
@camaraproject/device-status_maintainers connected network subscriptions aligned.

What strategy should we follow for non-mandatory errors? What errors do you think we should keep/remove?

akoshunyadi:
For the subscription APIs there is quite much overlapping with #250. Maybe we should work on the subscription APIs there and only on the retrieve APIs here. @fernandopradocabrillo what do you think?

fernandopradocabrillo:
> For the subscription APIs there is quite much overlapping with #250. Maybe we should work on the subscription APIs there and only on the retrieve APIs here. @fernandopradocabrillo what do you think?

@akoshunyadi agree, it made sense earlier but now there are too many conflicts, I'll remove the subscription changes and leave only the retrieve yamls

code/API_definitions/connected-network-type.yaml
```diff
@@ -279,142 +292,269 @@ components:
...
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 429
+                  code:
+                    enum:
+                      - QUOTA_EXCEEDED
+                      - TOO_MANY_REQUESTS
+          examples:
```
bigludo7, line 498: We have to align  message fieldd and description for Error 429
Defined in Commonalities https://github.com/camaraproject/Commonalities/pull/390

```
          examples:
            GENERIC_429_QUOTA_EXCEEDED:
              description: Request is rejected due to exceeding a business quota limit
              value:
                status: 429
                code: QUOTA_EXCEEDED
                message: Out of resource quota.
            GENERIC_429_TOO_MANY_REQUESTS:
              description: Access to the API has been temporarily blocked due to rate or spike arrest limits being reached
              value:
                status: 429
                code: TOO_MANY_REQUESTS
                message: Rate limit reached.
```

code/API_definitions/device-reachability-status.yaml
```diff
@@ -465,7 +531,17 @@ components:
...
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 429
+                  code:
+                    enum:
+                      - QUOTA_EXCEEDED
+                      - TOO_MANY_REQUESTS
```
bigludo7, line 544: Same - probably easy to align for new 429 message.

code/API_definitions/device-roaming-status.yaml
```diff
@@ -480,7 +544,17 @@ components:
...
+            allOf:
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 429
+                  code:
+                    enum:
+                      - QUOTA_EXCEEDED
```
bigludo7, line 556: Same

bigludo7 (commented):
@fernandopradocabrillo  - look good for me except that we do not have the 'new' 429 message. Let me know if you want to handle in this one or in other PR.

Should be (as defined in Defined in Commonalities https://github.com/camaraproject/Commonalities/pull/390 )

```
          examples:
            GENERIC_429_QUOTA_EXCEEDED:
              description: Request is rejected due to exceeding a business quota limit
              value:
                status: 429
                code: QUOTA_EXCEEDED
                message: Out of resource quota.
            GENERIC_429_TOO_MANY_REQUESTS:
              description: Access to the API has been temporarily blocked due to rate or spike arrest limits being reached
              value:
                status: 429
                code: TOO_MANY_REQUESTS
                message: Rate limit reached.
```

code/API_definitions/connected-network-type.yaml
```diff
@@ -279,142 +292,269 @@ components:
...
+            allOf:
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 401
+                  code:
+                    enum:
+                      - UNAUTHENTICATED
```
eric-murray, line 336: There is an open [issue](https://github.com/camaraproject/Commonalities/issues/368) in Commonalities to discuss the future of `UNAUTHENTICATED` / `AUTHENTICATION_REQUIRED`. I'm happy just to have `UNAUTHENTICATED` for now, but let's not forget to review again after Commonalities have made a decision.

code/API_definitions/connected-network-type.yaml
```diff
@@ -279,142 +292,269 @@ components:
...
+            allOf:
+              - $ref: "#/components/schemas/ErrorInfo"
+              - type: object
+                properties:
+                  status:
+                    enum:
+                      - 404
+                  code:
+                    enum:
+                      - NOT_FOUND
```
eric-murray, line 386: Do we have a use for `NOT_FOUND`? My preference would be to remove it.

code/API_definitions/connected-network-type.yaml
fernandopradocabrillo: I'm also happy with UNAUTHENTICATED

code/API_definitions/connected-network-type.yaml
```diff
@@ -279,142 +292,269 @@ components:
...
+                  code:
+                    enum:
+                      - CONFLICT
+          examples:
+            GENERIC_409_CONFLICT:
+              description: Duplication of an existing resource
+              value:
+                status: 409
+                code: CONFLICT
+                message: A specified resource duplicate entry found.
```
eric-murray, line 425: 409 Conflict is not used by this API, so I'd suggest to remove the error definition itself completely

eric-murray:
@akoshunyadi
> > @camaraproject/device-status_maintainers shouldn't the POST /subscription endpoint also have the 404 IDENTIFIER_NOT_FOUND error?
> 
> You mean for missing phone number? For that case I think 422 - MISSING_IDENTIFIER should be returned

IDENTIFIER_NOT_FOUND is the error to return when the API consumer specifies a valid identifier (e.g. MSISDN), but that MSISDN is not managed by the API provider (i.e. the MSISDN is managed by another network operator). In an aggregated world, hopefully this will never happen, but it might if the API consumer is directly consuming the API from the network operator.

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -400,7 +397,7 @@ components:
         subscriptionExpireTime:
           type: string
           format: date-time
-          example: "2023-01-17T13:18:23.682Z"
+          example: 2023-01-17T13:18:23.682Z
```
eric-murray, line 400: The "quotation" marks were added because of [this](https://github.com/camaraproject/DeviceStatus/issues/238) issue. So they all need to stay for date-time examples.

code/API_definitions/device-reachability-status-subscriptions.yaml
```diff
@@ -164,9 +164,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Subscription"
-              examples:
-                subscription-active:
-                  $ref: "#/components/examples/SUBSCRIPTION_ACTIVE"
```
eric-murray, line 169: This example needs to stay

code/API_definitions/connected-network-type.yaml
fernandopradocabrillo: done, thanks!

code/API_definitions/connected-network-type.yaml
fernandopradocabrillo: Agree, I'll remove it

fernandopradocabrillo:
> @fernandopradocabrillo Whilst reviewing, I noted that your fork does not appear to be synchronised to the latest version of the repository. You have clobbered all the recent changes to `device-reachability-status-subscriptions.yaml`, for example.
> 
> Can you sync your fork and resolve any outstanding conflicts first before I continue my review? Thanks.

I was trying to remove all the changes I included for the subscriptions since a new PR was raised duplicating them and including new ones, but it has been a mess so yeah, I'll simply pull the changes from main and try to sync

fernandopradocabrillo:
> @akoshunyadi
> 
> > > @camaraproject/device-status_maintainers shouldn't the POST /subscription endpoint also have the 404 IDENTIFIER_NOT_FOUND error?
> > 
> > 
> > You mean for missing phone number? For that case I think 422 - MISSING_IDENTIFIER should be returned
> 
> IDENTIFIER_NOT_FOUND is the error to return when the API consumer specifies a valid identifier (e.g. MSISDN), but that MSISDN is not managed by the API provider (i.e. the MSISDN is managed by another network operator). In an aggregated world, hopefully this will never happen, but it might if the API consumer is directly consuming the API from the network operator.

maybe we can move this discussion to PR #250 and include the error there if it is needed. In theory it could happen that in a 2-legged scenario, the client specifies a valid msisdn that doesn't belong to the operator

code/API_definitions/connected-network-type.yaml
akoshunyadi: I agree, we need only IDENTIFIER_NOT_FOUND

code/API_definitions/device-reachability-status.yaml
```diff
@@ -427,37 +472,64 @@ components:
...
                 status: 422
-                code: DEVICE_NOT_APPLICABLE
-                message: The service is not available for the provided device.
+                code: UNNECESSARY_IDENTIFIER
+                message: The device is already identified by the access token.
             GENERIC_422_UNABLE_TO_PROVIDE_REACHABILITY_STATUS:
               value:
                 status: 422
                 code: UNABLE_TO_PROVIDE_REACHABILITY_STATUS
                 message: Network issue - Unable to provide reachability status
```
akoshunyadi, line 532: Is this case really a 422 Unprocessable Content ? If the request is incorrect, we should tell which parameter is wrong. But this one looks like a 5xx error, maybe 503, to me.

code/API_definitions/device-roaming-status.yaml
```diff
@@ -430,48 +467,82 @@ components:
...
+              description: An explicit identifier is provided when a device or phone number has already been identified from the access token
+              value:
+                status: 422
+                code: UNNECESSARY_IDENTIFIER
+                message: The device is already identified by the access token.
             GENERIC_422_UNABLE_TO_PROVIDE_ROAMING_STATUS:
               value:
                 status: 422
                 code: UNABLE_TO_PROVIDE_ROAMING_STATUS
                 message: Network issue - Unable to provide roaming status
```
akoshunyadi, line 545: Same as above: Is this case really a 422 Unprocessable Content ? If the request is incorrect, we should tell which parameter is wrong. But this one looks like a 5xx error, maybe 503, to me.

code/API_definitions/connected-network-type.yaml
```diff
@@ -136,15 +140,13 @@ paths:
...
+          $ref: "#/components/responses/Generic404"
         "422":
-          $ref: "#/components/responses/ConnectedStatusUnprocessableEntity422"
-        "500":
-          $ref: "#/components/responses/Generic500"
-        "503":
-          $ref: "#/components/responses/Generic503"
+          $ref: "#/components/responses/Generic422"
+        "429":
+          $ref: "#/components/responses/Generic429"
```
akoshunyadi, line 149: This API doesn't have 406 and 415 as the other 2. I think for consistency reason the responses should be same for all 3 APIs. I'm not sure if we need those though...

code/API_definitions/device-reachability-status.yaml
fernandopradocabrillo: isn't this error for when it is not possible to return the real-time status of the device? I'm not sure it should be a 5XX, that would be telling the client to not retry the request with the same data, but this might be a temporary error

akoshunyadi: if only the referenced device is affected, not generally, then I would add the "device" to the case. e.g.
 GENERIC_422_UNABLE_TO_PROVIDE_REACHABILITY_STATUS_FOR_DEVICE:
                code: UNABLE_TO_PROVIDE_REACHABILITY_STATUS_FOR_DEVICE

eric-murray: > Is this case really a 422 Unprocessable Content ? If the request is incorrect, we should tell which parameter is wrong. But this one looks like a 5xx error, maybe 503, to me.

This is the error when the API consumer provides both a 3-legged token **and** explicit device identifier. The API provider cannot confirm whether or not the two identifiers match, as this would allow the API consumer to confirm the identity associated with the token, which they might otherwise not know (so a privacy issue).

So the API consumer must provide one or the other, but not both. See [here](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#appendix-a-normative-infodescription-template-for-when-user-identification-can-be-from-either-an-access-token-or-explicit-identifier) for a summary of this issue.

akoshunyadi: @bigludo7 @eric-murray @sachinvodafone Do we need this service specific code at all? Is there a use case for that?

code/API_definitions/connected-network-type.yaml
fernandopradocabrillo: I think that the tendency is to remove unused errors, maybe what we have to do is delete them from the other two apis

code/API_definitions/device-reachability-status.yaml
eric-murray: I agree this should be a 503. That assumes that the error is temporary and that normally a status can be returned for the device.

If there are devices for which a reachability status can never be returned, that is `422 SERVICE_NOT_APPLICABLE`.

code/API_definitions/device-roaming-status.yaml
eric-murray: Same comment as above:
- Temporary problem: 503
- Permanent problem: 422 SERVICE_NOT_APPLICABLE

code/API_definitions/connected-network-type.yaml
eric-murray: I agree. We should remove 415 and 406 from all APIs. Neither is required to be documented, and their use in these APIs is nothing out of the ordinary for those status codes, and not related to the use case.

bigludo7: Agree as well - we should remove 415 & 406.

code/API_definitions/device-reachability-status.yaml
bigludo7: I'm fine with @eric-murray proposal and not introduce a specific code.
Perhaps worth to ad a note in commonalities as 503 definition is "the server is currently unable to handle the request due to a temporary overload or scheduled maintenance" - here we have one additional case - we're not able to provide the service for a device we're supposed to be able to provide.

eric-murray: @bigludo7
Yes, agree. We can explicitly include 503 in the YAML with the "network issue" message as an example.

fernandopradocabrillo: Updated accordingly

code/API_definitions/device-roaming-status.yaml
fernandopradocabrillo: Updated

code/API_definitions/connected-network-type.yaml
fernandopradocabrillo: Done

akoshunyadi (approved):

maxl2287 (approved):
LGTM

maxl2287:
@eric-murray @sachinvodafone @bigludo7 as this PR is blocking the test-PRs, could you be so kind and review this again please?

Many thanks in advance.

sachinvodafone (approved):

eric-murray (approved):
I don't see any scenario that would result in `404 NOT_FOUND`, but we can revisit that at a later date

bigludo7 (approved):
LGTM

maxl2287:
@akoshunyadi @eric-murray can we merge this now?
I will update everything regarding the error models after this PR and #251 is merged during my #255.

Wdyt?

eric-murray:
Yes, we should merge this now, and then review other PRs.

I can merge if @akoshunyadi is unavailable

maxl2287:
Thanks! :)

maxl2287:
## Addiotional remark

This PR also fixes https://github.com/camaraproject/DeviceStatus/issues/254
Fixes #254

PR #247, opened 2025-01-23
Clarification for initialEvent for connected-network-type-subscriptions

### LINKED ISSUE #222: Enhance API documentation

bigludo7:
#### What type of PR is this?

* documentation

Clarification for initialEvent for connected-network-type-subscriptionsns.yaml

Fixes #222 

#### Changelog input

```
 release-note
- documentation: Clarification for initialEvent for connected-network-type-subscriptions

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/connected-network-type-subscriptions.yaml
+++ b/code/API_definitions/connected-network-type-subscriptions.yaml
@@ -44,6 +44,18 @@ info:
       - the Access Token `sinkCredential` (optionally set by the requester) expiration time has been reached
       - the API server has to stop sending notification prematurely
 
+    **Note on combined usage of ``initialEvent`` and ``subscriptionMaxEvents``**:
+
+    If an event is triggered following ``initialEvent`` set to true,
+    this event will be counted towards ``subscriptionMaxEvents`` (if provided).
+
+
+    **Clarification on ``initialEvent`` & ``event-type`` behaviour:**
+
+    When ``initialEvent`` is set to ``true`` in the subscription request, a notification has to be sent immediatly to provide the connected network type of the device targeted.
+    If the device is not connected, `UNKNOWN` is sent as ``connectedNetworkType`` in this notification.
+
+    
     ### Notifications callback
 
     This endpoint describes the event notification received on subscription listener side when the event occurred.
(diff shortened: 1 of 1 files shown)


code/API_definitions/connected-network-type-subscriptions.yaml
```diff
@@ -44,6 +44,17 @@ info:
...
+
+    If an event is triggered following ``initialEvent`` set to true,
+    this event will be counted towards ``subscriptionMaxEvents`` (if provided).
+
+
+    **Clarification on ``initialEvent`` & ``event-type`` behaviour:**
+
+    When ``initialEvent`` is set to ``true`` in the subscription request, a notification has to be sent immediatly to provide the connected network type of the device targeted.
+    If the device is not connected, `UNKNOWN` is sent as ``connectedNetworkType`` in this notification.
+
```
akoshunyadi, line 57: - typo in "immediatly"
- UNKNOWN is sent if connection [technology] can not be determined.

bigludo7: For the second point @akoshunyadi. If you subscribe with initalEvent set to true and the device si not connected (suppose the device is shut off) then we did trigger any event then?

I'm fine with that but looking for feedback.
+ @eric-murray  & @sachinvodafone  to get their view.

akoshunyadi: I wanted to point out the UNKNOWN is defined as "can not be determined" and not "not connected". Not connected is one of the cases when it can't determined, but there are other use cases for that too.
To your question: I think we should send the initial event, even if the status is unknown.

bigludo7: OK got it. 
If the status can not be determined we send an inital event valued to UNKNOWN. If the mobile is not connected no event triggered. right? I update

akoshunyadi (approved):

PR #262, opened 2025-02-28
Make lastStatusTime mandatory in 200 responses of the status APIs

### LINKED ISSUE #138: How to respond if queried device does not have an active session?
**Problem description**

There may be cases where the queried device for connectivity or roaming does not have an active session.  An appropriate API response needs to be agreed to ensure same behavior by multiple API providers.
**Possible evolution**

- An error response may be returned with an appropriate message.  

- A success message with a dedicated property maybe used to indicate device is not online

**Alternative solution**

**Additional context**

see: https://github.com/camaraproject/Commonalities/issues/127#issuecomment-2045370296

eric-murray:
#### What type of PR is this?
* enhancement/feature

`lastStatusTime` is an optional parameter in the 200 responses of the status APIs. But there is no good reason not to include this parameter. Otherwise, there is the risk of ambiguity as to whether the returned status is still valid. It should be mandatory for any status that is not current, and the simplest way to enforce that is to make it mandatory for all responses.

Fixes #138 

Could be part of Spring'25 meta-release if approved and merged today

#### Changelog input

```
 release-note
 - Make lastStatusTime mandatory in 200 responses of the status APIs
```

#### Additional documentation 
None

### DIFF
--- a/code/API_definitions/connected-network-type.yaml
+++ b/code/API_definitions/connected-network-type.yaml
@@ -173,6 +173,7 @@ components:
     ConnectedNetworkTypeResponse:
       type: object
       required:
+        - lastStatusTime
         - connectedNetworkType
       properties:
         lastStatusTime:
--- a/code/API_definitions/device-reachability-status.yaml
+++ b/code/API_definitions/device-reachability-status.yaml
@@ -179,6 +179,7 @@ components:
     ReachabilityStatusResponse:
       type: object
       required:
+        - lastStatusTime
         - reachable
       properties:
         lastStatusTime:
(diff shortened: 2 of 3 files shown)


sachinvodafone (approved):

akoshunyadi:
Is this PR meant for Spring25?

bigludo7 (approved):
LGTM

fernandopradocabrillo:
Changing an optional reponse property into required is a breaking change, we'll need to have this into consideration for Fall25, speacially for roaming and reachability that are already in stable v1.0

eric-murray:
> Changing a reponse property from optional to required is a breaking change, we'll need to have this into consideration for Fall25, specially for roaming and reachability which are already in stable v1.0

It's not a breaking change for the client, so a "v1" client will still work, even with this change

